PR #3, opened 2016-01-04
bump geo version

samlecuyer:
We need to bump the version so the latest versions of rust-gdal, rust-geo, and rust-proj all play nice together.

PR #4, opened 2017-08-09
Bump dependency version

mthh:
This PR (also related to #3) updates the version number of the `geo` crate and adds the `num_traits` crate as a dependency, so this should allow for better integration with the `geo` crate.

By the way, I don't know what is the best pattern to follow in this kind of cases (so I don't particularly advocate this solution), but is it really necessary to use the type `geo::Point` as input / output for the `Proj::project` method?
This involves adding the "geo" crate to a project only to use this function (while it might take as input / output something like `(f64, f64)` or `Vec<f64>` ; that's what I did on [another branch](https://github.com/mthh/rust-proj/blob/withoutgeo/src/proj.rs#L56) for a project not using the `geo` crate for now).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: rust
 rust:
-  - 1.0.0
   - stable
   - beta
   - nightly
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,4 +13,4 @@ license = "Apache-2.0"
 [dependencies]
 geo ="^0.4"
 libc = "0.1.7"
-num-traits = "0.1.39"
+num-traits = "^0.1"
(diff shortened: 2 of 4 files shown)


mthh:
I guess we could also remove `rust 1.0.0` from travis configuration file.

frewsxcv:
Bumping to 0.2.0 since it's technically a breaking change since we bumped rust-geo

frewsxcv:
> but is it really necessary to use the type geo::Point as input / output for the Proj::project method?

not really, what you did is cool

frewsxcv:
> but is it really necessary to use the type geo::Point as input / output for the Proj::project method?

not really, what you did in your other branch is cool for removing the `geo` dependency

mthh:
Thanks!

PR #5, opened 2018-03-07
Updates to bindings and docs

urschrei:
- specify `proj.4`, now that `proj.5` has been released
- Update dependencies
- Replace use of tuples (not layout-safe for FFI) with `c_void`
- Add rudimentary inline docs
- Add example from tests to README

PR #6, opened 2018-03-08
Move to proj-sys and proj.4 v5

urschrei:
With the surprising release (after about 23 years) of proj.4 v5, and my blowing-the-cobwebs-off PR from yesterday (#5), I thought I'd publish a [`proj-sys`](https://crates.io/crates/proj-sys) crate (generated using `bindgen`), and use it as the basis for this.
Currently, the API contains a very breaking change, because proj.4 performs operations differently now: it distinguishes between _projection_ (conversions from geodetic to projected coordinates) and _conversion_ (conversions between projected coordinate systems, usually within the same datum / reference frame). The latter uses a concept proj.4 calls the [`pipeline`](http://proj4.org/operations/pipeline.html), which allows for multi-step operations to achieve almost any kind of conversion, as opposed to the old system, which only allowed the specification of `from` and `to` projections.
This change is mirrored in the rust-proj API: for simple (inverse) projection from/to geodetic coordinates, there's `project`, and for conversions and transformations there's `convert`.
I'm not 100% sold on the API, because it feels a bit less flexible than the old one: you could specify different destination projections, and re-use the `Proj` struct. Using the new API, you have to specify everything up front, but you have a great deal more flexibility in terms of what you can do with it (see the `conversion` example). The conversion function also returns a `Result` now, and errors are implemented using `Failure`.
<s>Anyway, this is currently WIP until I figure out how to detect errors using the new API.</s>

### DIFF
--- a/src/proj.rs
+++ b/src/proj.rs
@@ -91,6 +91,11 @@ mod test {
     use geo::Point;
     use super::Proj;
 
+    fn assert_almost_eq(a: f64, b: f64) {
+        let f: f64 = a / b;
+        assert!(f < 1.00001);
+        assert!(f > 0.99999);
+    }
     #[test]
     fn test_new_projection() {
         let wgs84 = "+proj=longlat +datum=WGS84 +no_defs";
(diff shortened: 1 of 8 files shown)


urschrei:
Update: error handling works now.

urschrei:
Ugh bindgen is choking on multiple definitions after I `included` the legacy API so I could look up errors. Hrmm.

src/proj.rs
```diff
@@ -94,36 +194,103 @@ mod test {
...
+            .convert(Point::new(4760096.421921, 3744293.729449))
+            .unwrap();
+        assert_almost_eq(t.x(), 1450880.29);
+        assert_almost_eq(t.y(), 1141263.01);
+    }
+    #[test]
+    #[should_panic]
+    // Test that instantiation fails wth bad proj string input
+    fn test_init_error() {
+        let _ = Proj::new("ugh").unwrap();
```
frewsxcv: alternatively

```rust
assert!(Proj::new("ugh").is_none());
```

frewsxcv:
@urschrei just gave you review permissions for the repo

urschrei:
bors r=frewsxcv

PR #14, opened 2019-05-01
Call proj_errno_reset before convert or project

eminence:
Without resetting errno, any previous errors will persist, causing
all future calls to convert() or project() to fail


urschrei:
Oh, good catch. Thanks!

urschrei:
Released on crates.io as 0.9.2

PR #17, opened 2019-09-04
Add bulk conversion

urschrei:


urschrei:
There may be a more efficient way of re-using the `points` slice, but if there is, it isn't straightforward: `Point<T>` and `PJ_COORD` are different sizes, so they can't be `transmute`d, meaning we _have_ to allocate in order to perform the `&mut [Point<T>] --> &mut [PJ_COORD]` transform.

PR #21, opened 2020-01-08
Normalise input and output coordinate order

urschrei:
When transforming between two known CRS, input and output
coordinate order is now normalised to Lat, Lon / Easting Northing.

Note that this normalisation step is only applied to conversion operations between two known CRS (e.g. those that call [proj_create_crs_to_crs](https://proj.org/development/reference/functions.html#c.proj_create_crs_to_crs)), as per the PROJ API.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,8 @@
 # Changes
 
+## 0.14.0
+* Normalise input and output coordinate order to Lat, Lon / Easting, Northing for conversions between known CRS (#21)
+
 ## 0.13.0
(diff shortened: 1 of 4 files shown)


urschrei:
It's possible that this might cause breakage for some users, but:
- It's ultimately better than asking them to work with `Point` or `Coordinate` structs that are definitionally wrong because they're using flipped values for `x` and `y`
- We can't `fix` input and output behind the scenes because PROJ doesn't provide details of coordinate order for a given CRS.

PR #27, opened 2020-04-02
Updated Cargo.toml to use the latest version of 'geo-types'

dandxy89:
First potential contribution to a Rust library.

I've been using the geo, geojson and now the proj crate but I've found there is a conflict of version. Proj is using an older version as I'm getting an error which says `perhaps two different versions of crate geo_types are being used?`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,8 @@
 # Changes
+
+## 0.16.0
+* Update to geo-types v0.5.0
+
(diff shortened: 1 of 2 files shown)


urschrei:
Oops, sorry!

urschrei:
Published to crates.io!

dandxy89:
Thanks @urschrei much appreciated!

PR #31, opened 2020-06-20
More robust error handling

urschrei:
This is a breaking change (`info()` and `def()` now return `Result`, hence the semver bump), but this PR removes all instances of `unwrap()` in the crate. Other fun additions include making the error enum public for doc purposes, and annotating the errors where useful.


urschrei:
Bors r+

PR #34, opened 2020-07-15
Enable native network functionality without requiring libcurl

urschrei:
Uses `reqwest` and the network callback functionality provided by `libproj`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -19,29 +19,29 @@
 //! 1. If you don't require additional grids or other customisation:
 //!     - Call `Proj::new` or `Proj::new_known_crs`. This creates a transformation instance ([`Proj`](proj/struct.Proj.html))
 //! 2. If you require a grid for the transformation you wish to carry out, or you need to customise the search path or the grid endpoint:
-//!     - Create a new [`TransformBuilder`](proj/struct.TransformBuilder.html) by calling `TransformBuilder::new()`. It may be modified to enable network downloads, disable the grid, cache or modify search paths;
-//!     - Call [`TransformBuilder.proj()`](proj/struct.TransformBuilder.html#method.proj) or [`TransformBuilder.transform_known_crs()`](proj/struct.TransformBuilder.html#method.transform_known_crs). This creates a transformation instance (`Proj`)
+//!     - Create a new [`ProjBuilder`](proj/struct.ProjBuilder.html) by calling `ProjBuilder::new()`. It may be modified to enable network downloads, disable the grid, cache or modify search paths;
+//!     - Call [`ProjBuilder.proj()`](proj/struct.ProjBuilder.html#method.proj) or [`ProjBuilder.transform_known_crs()`](proj/struct.ProjBuilder.html#method.transform_known_crs). This creates a transformation instance (`Proj`)
 //!
 //! **Note**:
 //!
-//! 1. Both `TransformBuilder` and `Proj` implement the [`Info`](proj/trait.Info.html) trait, which can be used to get information about the current state of the `PROJ` instance;
-//! 2. `Proj::new()` and `TransformBuilder::proj()` have the same signature;
-//! 3. `Proj::new_known_crs()` and `TransformBuilder::transform_known_crs()` have the same signature.
+//! 1. Both `ProjBuilder` and `Proj` implement the [`Info`](proj/trait.Info.html) trait, which can be used to get information about the current state of the `PROJ` instance;
+//! 2. `Proj::new()` and `ProjBuilder::proj()` have the same signature;
+//! 3. `Proj::new_known_crs()` and `ProjBuilder::transform_known_crs()` have the same signature.
 //!
 //! ## Network, Cache, and Search Path Functionality
 //!
 //! ### Grid File Download
 //! `proj` supports [network grid download](https://proj.org/usage/network.html) functionality.
 //! Network access is **disabled** by default, and
-//! can be activated by passing a `true` `bool` to [`enable_network()`](proj/struct.Context.html#method.enable_network).
+//! can be activated by passing a `true` `bool` to [`enable_network()`](proj/struct.ProjBuilder.html#method.enable_network).
 //! Network functionality status can be queried with
 //! `network_enabled`, and the download endpoint can be queried and set using `get_url_endpoint` and `set_url_endpoint`.
 //!
 //! #### Grid File Cache
-//! Up to 300 mb of downloaded grids are cached to save bandwidth: This cache can be enabled or disabled using [`grid_cache_enable`](proj/struct.Context.html#method.grid_cache_enable).
+//! Up to 300 mb of downloaded grids are cached to save bandwidth: This cache can be enabled or disabled using [`grid_cache_enable`](proj/struct.ProjBuilder.html#method.grid_cache_enable).
 //!
 //! ### Search Path Modification
-//! The path used to search for resource files can be modified using [`set_search_paths`](proj/struct.Context.html#method.set_search_paths)
+//! The path used to search for resource files can be modified using [`set_search_paths`](proj/struct.ProjBuilder.html#method.set_search_paths)
 //!
 //!
 //! # Requirements
@@ -81,7 +81,7 @@ mod network;
 mod proj;
 
 pub use crate::proj::Area;
-pub use crate::proj::TransformBuilder;
+pub use crate::proj::ProjBuilder;
 pub use crate::proj::Info;
 pub use crate::proj::Proj;
 pub use crate::proj::ProjError;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -20,13 +20,13 @@
 //!     - Call `Proj::new` or `Proj::new_known_crs`. This creates a transformation instance ([`Proj`](proj/struct.Proj.html))
 //! 2. If you require a grid for the transformation you wish to carry out, or you need to customise the search path or the grid endpoint:
 //!     - Create a new [`ProjBuilder`](proj/struct.ProjBuilder.html) by calling `ProjBuilder::new()`. It may be modified to enable network downloads, disable the grid, cache or modify search paths;
-//!     - Call [`ProjBuilder.proj()`](proj/struct.ProjBuilder.html#method.proj) or [`ProjBuilder.transform_known_crs()`](proj/struct.ProjBuilder.html#method.transform_known_crs). This creates a transformation instance (`Proj`)
+//!     - Call [`ProjBuilder.proj()`](proj/struct.ProjBuilder.html#method.proj) or [`ProjBuilder.proj_known_crs()`](proj/struct.ProjBuilder.html#method.proj_known_crs). This creates a transformation instance (`Proj`)
 //!
 //! **Note**:
 //!
 //! 1. Both `ProjBuilder` and `Proj` implement the [`Info`](proj/trait.Info.html) trait, which can be used to get information about the current state of the `PROJ` instance;
 //! 2. `Proj::new()` and `ProjBuilder::proj()` have the same signature;
-//! 3. `Proj::new_known_crs()` and `ProjBuilder::transform_known_crs()` have the same signature.
+//! 3. `Proj::new_known_crs()` and `ProjBuilder::proj_known_crs()` have the same signature.
 //!
 //! ## Network, Cache, and Search Path Functionality
 //!
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -20,12 +20,12 @@
 //!     - Call `Proj::new` or `Proj::new_known_crs`. This creates a transformation instance ([`Proj`](proj/struct.Proj.html))
 //! 2. If you require a grid for the transformation you wish to carry out, or you need to customise the search path or the grid endpoint:
 //!     - Create a new [`TransformBuilder`](proj/struct.TransformBuilder.html) by calling `TransformBuilder::new()`. It may be modified to enable network downloads, disable the grid, cache or modify search paths;
-//!     - Call [`TransformBuilder.transform()`](proj/struct.TransformBuilder.html#method.transform) or [`TransformBuilder.transform_known_crs()`](proj/struct.TransformBuilder.html#method.transform_known_crs). This creates a transformation instance (`Proj`)
+//!     - Call [`TransformBuilder.proj()`](proj/struct.TransformBuilder.html#method.proj) or [`TransformBuilder.transform_known_crs()`](proj/struct.TransformBuilder.html#method.transform_known_crs). This creates a transformation instance (`Proj`)
 //!
 //! **Note**:
 //!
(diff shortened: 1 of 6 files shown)


urschrei:
Potential reviewers: this is a wrapper for a cpp library, so it already contains a great deal of unsafe code. However, since this crate also provides a lot of optional functionality to `geo`, and the new functionality involves a lot (relatively speaking) of raw pointer manipulation, I thought I'd ask for some additional feedback, since I don't work with them much.

src/proj.rs
```diff
@@ -86,11 +107,71 @@ fn area_set_bbox(parea: *mut proj_sys::PJ_AREA, new_area: Option<Area>) {
...
+pub fn enable_network(enable: bool) -> Result<(), ProjError> {
+    if enable {
+        let _ = match set_network_callbacks() {
+            1 => Ok(1),
+            _ => Err(ProjError::Network),
+        }?;
+    }
+    let enable = if enable { 1 } else { 0 };
+    let dctx: *mut PJ_CONTEXT = ptr::null_mut();
+    unsafe { proj_context_set_enable_network(dctx, enable) };
```
michaelkirk, line 124: should we verify the return? 

https://proj.org/development/reference/functions.html#_CPPv431proj_context_set_enable_networkP10PJ_CONTEXTi

> returns TRUE if network access has been enabled

```
int proj_context_set_enable_network(PJ_CONTEXT *ctx, int enable) {
    if (ctx == nullptr) {
        ctx = pj_get_default_ctx();
    }
    // Load ini file, now so as to override its network settings
    pj_load_ini(ctx);
    ctx->networking.enabled_env_variable_checked = true;
    ctx->networking.enabled = enable != FALSE;
#ifdef CURL_ENABLED
    return ctx->networking.enabled;
#else
    return ctx->networking.enabled &&
           ctx->networking.open != NS_PROJ::no_op_network_open;
#endif
}
```

It seems almost but not quite redundant, e.g. if there were a situation where `ctx->networking.open != NS_PROJ::no_op_network_open` failed...

src/proj.rs
```diff
@@ -255,6 +297,52 @@ impl Proj {
...
+        let paths_c = individual
+            .iter()
+            .map(|str| CString::new(*str))
+            .collect::<Result<Vec<_>, std::ffi::NulError>>()?;
+        // …then to raw pointers
+        let paths_p: Vec<_> = paths_c.iter().map(|cstr| cstr.as_ptr()).collect();
+        // …then pass the slice of raw pointers as a raw pointer (const char* const*)
+        // We pass a null pointer as the context, as we want the search path to be
+        // available to all contexts
+        let dctx: *mut PJ_CONTEXT = ptr::null_mut();
```
michaelkirk, line 340: I'm not at all familiar with this crate - this PR is the first I've looked into it, so I could be off base, but I'm confused at this point...

Here's my understanding of PJ_CONTEXT:

In the c interface, all proj config and state is wrapped up into a proj context. 

You can create and manage a context, and pass it into the methods which get and set state. As a convenience, you can use the global default context by passing in a null ptr instead of your own context.

Before this PR, this method seemed to be the only place we use the default context. I'm not sure why we were doing that - my first impulse is that it's strange. It seems surprising to me that a method which is on an explicit instance would also affect another instance (the global default context). 

The new network methods added in this PR `network_enabled`, `enable_network(enable: bool) -> Result<(), ProjError>`, etc. now also operate on the default ctx. Why is that? 

I would have expected that they would be impl methods on an instance of Proj, which read and write only to their Proj context and not touch the global default context.

```
let my_proj = Proj::new(a,b,c);
my_proj.get_url_endpoint()
```

If we did need/want a way to manipulate the global default context, maybe best to do that through a static instance like 
```rust
Proj::default.get_url_endpoint()
```

But take that critique with a grain of salt - I've never used this lib directly. 🙃

src/network.rs
```diff
@@ -0,0 +1,361 @@
...
+    // Retries are exhausted to no avail, so give up with an error
+    if status >= 300 {
+        return Err(ProjError::DownloadError(
+            res.status().as_str().to_string(),
+            url,
+            retries,
+        ));
+    }
+    // Write the initial read length value into the pointer
+    let contentlength = res.content_length().ok_or(ProjError::ContentLength)? as usize;
```
michaelkirk, line 161: 🐫

src/proj.rs
urschrei: Done!

src/proj.rs
```diff
@@ -65,7 +86,7 @@ impl Area {
 }
 
 /// Easily get a String from the external library
-fn _string(raw_ptr: *const c_char) -> Result<String, ProjError> {
+pub fn _string(raw_ptr: *const c_char) -> Result<String, ProjError> {
```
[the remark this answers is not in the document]
urschrei: Done!

src/proj.rs
```diff
@@ -595,6 +695,21 @@ mod test {
         assert_almost_eq(t.x(), 1450880.29);
         assert_almost_eq(t.y(), 1141263.01);
     }
+    // #[test]
```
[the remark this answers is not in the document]
urschrei: Done!

src/proj.rs
urschrei: Great, thank you for this!  
The issue here is one of my own making:
In order to create transformation / projection instances using the crate, you can call `Proj::new` / `Proj::new_known_crs`. Note that both of these methods first create a `ctx` – the implicit context creation was a design decision on my part because I thought it would be tedious to explicitly create e.g. a `Context` instance, _then_ call a projection / transformation method on it.  
However, modifying search paths, network download functionality etc requires a `ctx`, as you point out – changing network download and search paths has to be done _before_ the transformation / projection object is created, because their creation may require search path lookup or remote grid file download. But because there is now no access to a `ctx` independently of a `Proj` instance, these properties can't be modified for an existing instance: they have to be set before creation. So there are three choices:

1. Break the API by requiring a two-step process: context creation followed by transformation / projection specification which consumes the ctx, with modification to search paths and network functionality as an intermediate step;
2. Break the API by adding new arguments to `Proj::new` / Proj::new_known_crs`
3. Leave the API as-is currently, which is at odds to the `libproj` API – my feeling is that crate users don't generally want or need to control these properties for concurrently existing `Proj` instances, and the existing effects of modifying properties are explicitly detailed in the comments.

src/proj.rs
```diff
@@ -121,8 +121,10 @@ pub fn enable_network(enable: bool) -> Result<(), ProjError> {
     }
     let enable = if enable { 1 } else { 0 };
     let dctx: *mut PJ_CONTEXT = ptr::null_mut();
-    unsafe { proj_context_set_enable_network(dctx, enable) };
-    Ok(())
+    match unsafe { proj_context_set_enable_network(dctx, enable) } {
+        1 => Ok(1),
+        _ => Err(ProjError::Network),
```
michaelkirk, line 126: My understanding is that with this change we'll return an error when calling:

```
enabled_network(enable: false)
```

from `proj_context_set_enable_network`:
```
...
    return ctx->networking.enabled &&
           ctx->networking.open != NS_PROJ::no_op_network_open;
```

But was that the intention?

src/proj.rs
```diff
@@ -595,6 +697,22 @@ mod test {
...
+    // This test is disabled by default as it requires network access
+    // #[test]
+    // fn test_network() {
+    //     let from = "EPSG:4326";
+    //     let to = "EPSG:4326+3855";
+    //     // off by default
+    //     assert_eq!(network_enabled(), false);
+    //     // switch it on and disable cache for subsequent calls
+    //     grid_cache_set_enable(false);
+    //     enable_network(true).unwrap();
```
michaelkirk, line 709: 😬 I feel like you're going to regret asking me to review this, because I have so many questions, but...

Can you explain what this is test testing? 

If I turn off my network, the test fails, which seems reasonable.

But I was surprised I could make it pass by applying this patch:
```diff
         assert_eq!(network_enabled(), false);
         // switch it on and disable cache for subsequent calls
         grid_cache_set_enable(false);
-        enable_network(true).unwrap();
+        // enable_network(true).unwrap();
         let proj = Proj::new_known_crs(&from, &to, None).unwrap();
-        assert_eq!(network_enabled(), true);
+        // assert_eq!(network_enabled(), true);
         let t = proj.convert(Point::new(40.0, -80.0)).unwrap();
         assert_almost_eq(t.x(), 39.99999839);
         assert_almost_eq(t.y(), -79.99999807);
```

So to clarify, it seems like here we're testing the network configuration - but none of the computations in the test actually require network, is that right?

urschrei: No, don't worry, I asked you because I knew this would happen!
You're absolutely correct, I just concocted a transform - the main idea was to trigger a download. I'll figure out something unambiguous that makes the test's intent clear.

src/proj.rs
urschrei: Well, this is embarrassing. I'll revisit tomorrow post-caffeine.

urschrei: I think this is now correct…

src/proj.rs
urschrei: The new test (`test_network_enabled_conversion`) demonstrates different results for the same conversion operation, depending on whether a grid was downloaded by enabling network functionality.

urschrei:
Adding some additional context from a related conversation w @michaelkirk: https://github.com/urschrei/rust-proj/pull/1#pullrequestreview-452074217

src/proj.rs
```diff
@@ -159,9 +169,64 @@ impl Proj {
...
+}
+
+impl Info for TransformBuilder {
+    #[doc(hidden)]
+    fn ctx(&self) -> *mut PJ_CONTEXT {
+        self.ctx
+    }
+}
+
+impl TransformBuilder {
```
michaelkirk, line 200: Naming nit, but maybe this would be clearer as `ProjBuilder` rather than introducing the concept of a "Transform" as something which might be distinct from "Proj"

Take it or leave it though - it's also fine as is.

src/proj.rs
```diff
@@ -159,9 +169,64 @@ impl Proj {
...
+        }) {
+            // we asked to switch on: switched on
+            (1, 1) => Ok(1),
+            // we asked to switch off: switched off
+            (0, 0) => Ok(0),
+            // we asked to switch off, but it's still on
+            (0, 1) => Err(ProjError::Network),
+            // we asked to switch on, but it's still off
+            (1, 0) => Err(ProjError::Network),
+            // scrëm
```
michaelkirk, line 224: 😆

src/proj.rs
```diff
@@ -180,14 +245,171 @@ impl Proj {
...
+    /// **Note:** for projection operations, `definition` specifies
+    /// the **output** projection; input coordinates
+    /// are assumed to be geodetic in radians, unless an inverse projection is intended.
+    ///
+    /// For conversion operations, `definition` defines input, output, and
+    /// any intermediate steps that are required. See the `convert` example for more details.
+    ///
+    /// # Safety
+    /// This method contains unsafe code.
+    pub fn transform(mut self, definition: &str) -> Option<Proj> {
```
michaelkirk, line 326: another naming nit which you can feel free to take or leave: `proj` or `build_proj`

src/network.rs
```diff
@@ -0,0 +1,358 @@
...
+    res: &'a mut Response,
+    rb: RequestBuilder,
+    hvalue: &str,
+) -> Result<&'a Response, ProjError> {
+    let mut status = res.status().as_u16();
+    let mut retries = 0;
+    // Check whether something went wrong on the server, or if it's an S3 retry code
+    if res.status().is_server_error() || RETRY_CODES.contains(&status) {
+        // Start retrying: up to MAX_RETRIES
+        while res.status().is_server_error()
```
michaelkirk, line 75: So we'll retry infinitely if we keep getting server errors? 

My instinct would be to count server errors like other failures and only retry a limited number of times.

Something like `(res.status().is_server_error() || RETRY_CODES.contains(&status)) && retries <= MAX_RETRIES`

src/network.rs
```diff
@@ -0,0 +1,358 @@
...
+            || RETRY_CODES.contains(&status)
+            || retries <= MAX_RETRIES
+        {
+            retries += 1;
+            let wait = time::Duration::from_millis(get_wait_time_exp(retries as i32));
+            thread::sleep(wait);
+            let retry = rb.try_clone().ok_or(ProjError::RequestCloneError)?;
+            let with_range = retry.header("Range", hvalue).header("Client", CLIENT);
```
michaelkirk, line 83: Just a small cleanup possibility: maybe we could get rid of setting the range header here in this method and set it on the request builder before we pass it into this method, no big deal either way though.

src/network.rs
```diff
@@ -0,0 +1,358 @@
...
+}
+
+/// Network callback: get header value
+fn _network_get_header_value(
+    _: *mut PJ_CONTEXT,
+    handle: *mut PROJ_NETWORK_HANDLE,
+    header_name: *const c_char,
+    _: *mut c_void,
+) -> Result<*const c_char, ProjError> {
+    let lookup = _string(header_name)?.to_lowercase();
```
michaelkirk, line 249: curious - why do we `to_lowercase` here? 

Is it a Reqwest thing that all headers are lowercased?

urschrei: Yep – this actually took me a solid day to figure out, because the download operation would error out after `network_open` had returned, for no discernible reason, even though it had retrieved data and written headers. Turns out that Reqwest (or possibly Hyper?) lowercases the headers. Headers are supposed to be case-insensitive, so this is a gotcha in the Hyper `HeaderMap` implementation: `get()` is case-sensitive, and maybe it shouldn't be.

src/network.rs
urschrei: I've just altered the retry condition on line 77  – we should exit the enclosing `if` on line 73 after 1 or `MAX_RETRIES` loops, failing on line 97 in case anything is still wrong.

src/lib.rs
```diff
@@ -1,19 +1,49 @@
...
+//! 1. Both `TransformBuilder` and `Proj` implement the [`Info`](proj/trait.Info.html) trait, which can be used to get information about the current state of the `PROJ` instance;
+//! 2. `Proj::new()` and `TransformBuilder::transform()` have the same signature;
+//! 3. `Proj::new_known_crs()` and `TransformBuilder::transform_known_crs()` have the same signature.
+//!
+//! ## Network, Cache, and Search Path Functionality
+//!
+//! ### Grid File Download
+//! `proj` supports [network grid download](https://proj.org/usage/network.html) functionality.
+//! Network access is **disabled** by default, and
+//! can be activated by passing a `true` `bool` to [`enable_network()`](proj/struct.Context.html#method.enable_network).
```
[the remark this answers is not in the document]
urschrei: Oops, good catch.

urschrei:
@michaelkirk I've addressed all the nits and tweaked the error handling logic – the new names are definitely an improvement. Thank you for the thorough review; I'm a **lot** happier with this.

michaelkirk:
Nice work @urschrei!

PR #36, opened 2020-08-17
disable default features, like openssl, since we're using rust-tls

michaelkirk:
I'm still testing this locally, but wanted to make sure you saw it @urschrei to avoid duplicated efforts.


urschrei:
Guess who just opened https://github.com/georust/proj/pull/37 🤦‍♂️

PR #37, opened 2020-08-17
Remove default features from reqwest

urschrei:
These include dependencies which require OpenSSL, which we wish to avoid.

### DIFF
--- a/src/proj.rs
+++ b/src/proj.rs
@@ -776,10 +776,9 @@ mod test {
     fn assert_almost_eq(a: f64, b: f64) {
         let f: f64 = a / b;
         assert!(f < 1.00001);
-        assert!(f > 0.99999);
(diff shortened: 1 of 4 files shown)


src/proj.rs
```diff
@@ -776,10 +776,9 @@ mod test {
     fn assert_almost_eq(a: f64, b: f64) {
         let f: f64 = a / b;
         assert!(f < 1.00001);
-        assert!(f > 0.99999);
+        assert!(f > 0.99999);    // This test should be disabled by default, as it requires network access
```
michaelkirk, line 779: I think moving this comment might have been unintentional?

urschrei: Hmm looks like rustfmt got confused. "We'll fix it in post"

urschrei: Fixed in 828d58e

PR #41, opened 2020-09-08
proj w/o geo

michaelkirk:
It seems like the potential users of proj would include people who don't need/want to opt into the whole `geo` ecosystem.

What would you think about having the proj interface be trait-based and having the geo-integration be a feature?

If this seems like a reasonable direction, before merging I'd like to update the docs to show:

1. how you can use this with the greater `geo` ecosystem
2. how you can `impl` your own point type if you're not using geo.
3. how you can use the built-in `(Float, Float)` impl

WDYT @urschrei?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ edition = "2018"
 
 [dependencies]
 proj-sys = { version = "0.18.3", path = "proj-sys" }
+geo-types = { version = "0.6", optional = true }
 libc = "0.2.62"
 num-traits = "0.2.8"
 thiserror = "1.0.4"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,6 @@ edition = "2018"
 
 [dependencies]
 proj-sys = { version = "0.18.3", path = "proj-sys" }
-geo-types ="0.6.0"
 libc = "0.2.62"
 num-traits = "0.2.8"
 thiserror = "1.0.4"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,5 +32,5 @@ network = ["reqwest"]
 assert_approx_eq = "1.1.0"
 
 [package.metadata.docs.rs]
-features = [ "proj-sys/nobuild", "network" ]
+features = [ "proj-sys/nobuild", "network", "geo-types" ]
 rustdoc-args = ["--cfg", "docsrs"]
--- a/README.md
+++ b/README.md
@@ -13,10 +13,10 @@ projected coordinate systems. The PROJ [documentation](https://proj.org/operatio
 explains the distinction between these operations in more detail.
 
 This crate depends on [`libproj v7.1.x`](https://proj.org), accessed via the
-[`proj-sys`](proj-sys) crate. By default, `proj-sys` will try to find a pre-existing
-installation of libproj on your system. If an appropriate version of libproj cannot be found,
-the build script will attempt to build `libproj` from source. You may specify a from-source
-build with the [`bundled_proj` feature](#feature-flags).
+[`proj-sys`](https://docs.rs/proj-sys) crate. By default, `proj-sys` will try to find a
+pre-existing installation of libproj on your system. If an appropriate version of libproj
+cannot be found, the build script will attempt to build libproj from source. You may specify a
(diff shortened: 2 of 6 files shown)


src/geo_types.rs
```diff
@@ -0,0 +1,53 @@
...
+/// let from = "EPSG:2230";
+/// let to = "EPSG:26946";
+/// let nad_ft_to_m = Proj::new_known_crs(&from, &to, None).unwrap();
+/// let result = nad_ft_to_m
+///     .convert(Coordinate { x: 4760096.421921f64, y: 3744293.729449f64 })
+///     .unwrap();
+/// assert_approx_eq!(result.x, 1450880.29f64, 1.0e-2);
+/// assert_approx_eq!(result.y, 1141263.01f64, 1.0e-2);
+/// ```
+impl<T: crate::proj::CoordinateType> crate::Point<T> for geo_types::Coordinate<T> {
```
michaelkirk, line 16: Note - this PR would also let you do Proj stuff with Coords, not just points.

michaelkirk: Now that I'm typing it out - we could take this even further and have a `Projectable` trait or somesuch that just required implementing a Iter<Item=proj::Point<T>>, then we could get easy projection of all the geo-type's. 🤔

michaelkirk:
We could make the `geo-types` feature a default feature to make this a non-breaking change.

urschrei:
I'm in favour of this so long as we write clear documentation around it. In general, I find it tricky to present feature interaction in a clear and concise way, but we needn't let that stop us…

src/geo_types.rs
michaelkirk: I'm not going to pursue the Iter scenario in this PR, but would like to do it in some followup work.

README.md
```diff
@@ -22,10 +22,7 @@ By default, this crate depends on a pre-built `libproj`, so PROJ v7.1.x must be
 
 ## Convert from [NAD 83 US Survey Feet](https://epsg.io/2230) to [NAD 83 Meters](https://epsg.io/26946) Using EPSG Codes
 ```rust
```
michaelkirk, line 24: The README contains lots of snippets from the primary docs. Since it can be tedious and error prone to copy changes back and forth between them, I was hoping we could review just the primary docs, and once they're in good shape I can update the README before a final review.

michaelkirk (commented):
Ok, with the caveat of the README (see comment) I think this is ready for review. Can you take a look @urschrei?

urschrei (approved):
This lgtm, let's merge!

urschrei:
I think this probably just needs a rebase now that we've switched to GH Actions?

michaelkirk:
Ok! Rebased and addressed conflicts in the docs. 

Note that I'm (almost) wholesale copying the contents of README.md from the doc comments in src/lib.rs - the only divergence is the GH actions badge.

michaelkirk:
bors try

Cargo.toml
```diff
@@ -14,7 +14,7 @@ edition = "2018"
 
 [dependencies]
 proj-sys = { version = "0.18.3", path = "proj-sys" }
-geo-types ="0.6.0"
+geo-types = { version = "0.6", optional = true }
```
michaelkirk, line 17: I snuck this in - there wasn't a reason for us to peg the patch version of geo-types was there?

michaelkirk:
bors try

I'm not sure why - but [bors crashed](https://app.bors.tech/repositories/42/log)... let's try again.

michaelkirk:
bors r=urschrei

michaelkirk:
bors cancel

michaelkirk:
🤦 booooorrrsss

Oh well, I just wanted to fix a little typo in the README and squash the commits, but bors will do what bors wants.

michaelkirk:
Oh, actually that was my bad. I should have said `bors r-` instead of `bors cancel`. Sorry for doubting you bors 😬.

urschrei:
We have always known bots and people couldn't live in harmony. It's been a good run. (Thank you for doing this!)

michaelkirk:
wth bors

michaelkirk:
image
That eye. So piercing.

urschrei:
I feel judged.

PR #47, opened 2020-10-01
fix docsrs build with correct syntax for crate attribute

### LINKED ISSUE #46: Docs build error due to use of #[doc(cfg)] feature
https://docs.rs/crate/proj/0.20.4/builds/299220

This is…confusing, as the crate seems to be correctly set up as per https://stackoverflow.com/questions/61417452/how-to-get-a-feature-requirement-tag-in-the-documentation-generated-by-cargo-do

michaelkirk:
Fixes #46 

... I hope.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -81,7 +81,8 @@
 //! assert_approx_eq!(result.y(), 1141263.01f64, 1.0e-2);
 //! ```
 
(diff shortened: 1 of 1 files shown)


urschrei:
Only one way to find out 😤

bors r+

PR #49, opened 2020-10-08
dupe: proj w/o geo (trying to make bors happy)

michaelkirk:
Just trying to appease bors: this is a dupe of #41, but originating from a local branch instead of a branch in my fork.


michaelkirk:
bors try

PR #50, opened 2020-11-04
Add Proj::get_area_of_use

lnicola:


lnicola:
Added test.

src/proj.rs
```diff
@@ -508,6 +509,46 @@ impl Proj {
...
+                self.c_proj,
+                &mut out_west_lon_degree,
+                &mut out_south_lat_degree,
+                &mut out_east_lon_degree,
+                &mut out_north_lat_degree,
+                &mut out_area_name,
+            )
+        };
+        if res == 1
+            && out_west_lon_degree != -1000.0
```
michaelkirk, line 534: Does it really make sense to have a `Result::Ok(Option::None)`? 

As a consuming user, I'm having a hard time imagining calling code which needs to handle `Ok(None)` differently from `Err(ProjError)` - in other words, my intuition is that it'd be nice to instead express these `Ok(None)` cases as some kind of `Err(ProjError)`.

from the docs: https://proj.org/development/reference/functions.html#_CPPv420proj_get_area_of_useP10PJ_CONTEXTPK2PJPdPdPdPdPPKc

> TRUE in case of success, FALSE in case of error or if the area of use is unknown.

> out_west_lon_degree: Pointer to a double to receive the west longitude (in degrees). Or NULL. If the returned value is -1000, the bounding box is unknown.

So, I believe we're checking `-1000.0` here to see if the area is unknown.
But my read of the above documentation is that when the area is unknown, `res == 1` will never be true.

Would it better to do something like:
```
if res == 1 {
    # my understanding is that this should never be the case if `res == 1`
    debug_assert!(out_west_lon_degree != -1000.0);
    debug_assert!(out_south_lat_degree != -1000.0);
    debug_assert!(out_east_lon_degree != -1000.0);
    debug_assert!(out_north_lon_degree != -1000.0);
    ....
    Ok(...)
} else {
    if (out_west_lon_degree == -1000.0 || out_south_lat_degree == -1000.0 || ... ) {
        Err(ProjError::UknownArea)
    } else {
        Err(...)
    }
}
```

lnicola: I think so, see the implementation at https://github.com/OSGeo/PROJ/blob/master/src/iso19111/c_api.cpp#L1708.

michaelkirk: ah, source is king, thanks for the link!

So I see four cases in the method:
1. everything works - here's the bounding area -> `(Ok(Some))`
1. there was no error, but proj simply doesn't know the bounding area -> `Ok(None)`
1. libproj returned an error (`res==0`) -> `Ok(None)`
1. there was an error in our lib with string conversion (`_string`): -> `Err(ProjError)`

With your explanation (thanks!) I think it's reasonable to return `Ok(None)` for case 2.

I think a case could be made for having case 3 return some kind of `Err(ProjError)`, but I'm not sure if you can return a meaningful error.  Can you get something from `proj_errno`?

Feel free to push back - I'm less familiar with the proj crate than @urschrei.

lnicola: I updated the code to return `Ok(ProjError::UnknownArea)` in case 2. For 3, there doesn't seem to be a way to get an error string, based on my reading of the source:

```c++
    auto objectUsage = dynamic_cast<const ObjectUsage *>(obj->iso_obj.get());
    if (!objectUsage) {
        return false;
    }
    const auto &domains = objectUsage->domains();
    if (domains.empty()) {
        return false;
    }
```

(We get `false` and that's it)

lnicola:
Updated to use `MaybeUninit`.

src/proj.rs
lnicola: By the way, thank you for the review.

urschrei: Is there a meaningful distinction between Case 2 and Case 3, in terms of how one would proceed? I feel like in either case, returning an error is acceptable. HOWEVER, you can ask on the mailing list or the PROJ issue tracker if you want to put the matter beyond doubt: Even and Kristian are extremely helpful…

lnicola: I'll try to ask on the mailing list, but I'm starting to feel that it's quite possible that name is available, while the bounds aren't.

lnicola: Even says:

> It should be uncommon to have an area of use without its bounding box, but if 
you instanciate a CRS from WKT2 with a USAGE[AREA["foo]] node, you would be in 
that situation.
You'll get either all 4 bounds valid or at -1000, not a mix of valid and 
invalid values.

urschrei: But what is the difference between a valid value and -1000?

lnicola: My guess is that you can define projections that don't have any bounds. It probably won't happen for known ones, but it can be the case for custom-defined projections.

urschrei: OK, so Ok(None) isn't _nice_, but it does seem to be the correct return type in this instance. Maybe note the reasoning in the docstring, and go with it? What do you think @michaelkirk

lnicola: Reverted and also made name optional (since it can be missing).

michaelkirk: It's admittedly kind of a gnarly API, but I guess that's just because the method can fail in lots of different ways. 

I think, as it is, returning Ok(None) makes sense. :+1:

urschrei:
bors r=michaelkirk

PR #51, opened 2020-11-14
Make Area fields public

lnicola:


src/proj.rs
```diff
@@ -1114,11 +1114,10 @@ mod test {
         let proj = Proj::new("EPSG:3035").unwrap();
         let (area, name) = proj.area_of_use().unwrap();
         let area = area.unwrap();
-        let name = name.unwrap();
         assert_eq!(area.west, -35.58);
         assert_eq!(area.south, 24.6);
         assert_eq!(area.east, 44.83);
         assert_eq!(area.north, 84.17);
-        assert_eq!(name, "Europe - LCC & LAEA");
+        assert!(name.is_some());
```
lnicola, line 1121: `proj` is a bit of a moving target, I guess.

urschrei: I think it's the EPSG database in this case. Good call.

urschrei (approved):

PR #56, opened 2021-01-06
Fix array conversion

urschrei:
My suspicion is that creating the raw pointer in the function call was causing it to be dropped prematurely, resulting in garbage data. See [here](https://doc.rust-lang.org/std/ffi/struct.CString.html#method.as_ptr) for an analogous issue when getting a `CString` pointer. We avoid this by explicitly creating the pointer before the call. In addition, the doctest has been updated with a less complex example that doesn't use a proj string.

### DIFF
--- a/src/proj.rs
+++ b/src/proj.rs
@@ -791,17 +791,19 @@ impl Proj {
             })
             .collect::<Result<Vec<_>, ProjError>>()?;
         pj.shrink_to_fit();
+        // explicitly create the raw pointer to ensure it lives long enough
+        let mp = pj.as_mut_ptr();
         // Transformation operations are slightly different
         match op {
             Transformation::Conversion => unsafe {
                 proj_errno_reset(self.c_proj);
                 trans =
-                    proj_trans_array(self.c_proj, PJ_DIRECTION_PJ_FWD, pj.len(), pj.as_mut_ptr());
+                    proj_trans_array(self.c_proj, PJ_DIRECTION_PJ_FWD, pj.len(), mp);
                 err = proj_errno(self.c_proj);
             },
             Transformation::Projection => unsafe {
                 proj_errno_reset(self.c_proj);
-                trans = proj_trans_array(self.c_proj, inv, pj.len(), pj.as_mut_ptr());
+                trans = proj_trans_array(self.c_proj, inv, pj.len(), mp);
                 err = proj_errno(self.c_proj);
             },
         }
--- a/src/proj.rs
+++ b/src/proj.rs
@@ -698,16 +698,17 @@ impl Proj {
     /// use proj::{Proj, Coord};
     ///
     /// # use assert_approx_eq::assert_approx_eq;
-    /// let from = "EPSG:2230";
-    /// let to = "EPSG:26946";
+    /// // Convert from NAD83(NSRS2007) to NAD83(2011)
+    /// let from = "EPSG:4759";
+    /// let to = "EPSG:4317";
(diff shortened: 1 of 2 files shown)


urschrei:
Bors try

michaelkirk:
bors try

Interesting! Let's gather some stats here.

urschrei:
If your run passes, I'm gonna contrive a doctest with different CRS and coordinates. I have my doubts about the validity of the old Stereo70 conversion values, but there's less value in a doctest that just reproduces an existing known-good test.

urschrei:
bors try

urschrei:
These are CRS and coordinate values from https://geodesy.noaa.gov/NCAT/.

WDYT?

src/proj.rs
```diff
@@ -1086,13 +1089,15 @@ mod test {
...
         let to = "EPSG:26946";
         let ft_to_m = Proj::new_known_crs(&from, &to, None).unwrap();
-        let mut v = vec![
-            MyPoint::new(4760096.421921, 3744293.729449),
-            MyPoint::new(4760197.421921, 3744394.729449),
-        ];
-        ft_to_m.convert_array(&mut v).unwrap();
-        assert_almost_eq(v[0].x(), 1450880.2910605003f64);
-        assert_almost_eq(v[1].y(), 1141293.7960220212f64);
+        for _ in 0..100 {
```
michaelkirk, line 1092: I understand the motivation for this change based on the context we're in right now, but eventually it'll be weird and inexplicable. I'd say either revert the loop change or pull it into a separate test with an explanation of its provenance.

src/proj.rs
```diff
@@ -731,16 +732,16 @@ impl Proj {
...
+    /// let mut v = vec![
+    ///     (4760096.421921, 3744293.729449),
+    ///     (4760197.421921, 3744394.729449),
+    /// ];
+    /// ft_to_m.convert_array(&mut v).unwrap();
+    /// assert_approx_eq!(v[0].x(), 1450880.2910605003f64);
+    /// assert_approx_eq!(v[1].y(), 1141293.7960220212f64);
     /// ```
     ///
     /// # Safety
```
michaelkirk, line 747: *cough* 👀

src/proj.rs
```diff
@@ -731,16 +732,16 @@ impl Proj {
...
-    /// )
-    /// .unwrap();
-    /// // Geodetic -> Pulkovo 1942(58) / Stereo70 (EPSG 3844)
-    /// let mut v = vec![(0.436332, 0.802851)];
-    /// let t = stereo70.project_array(&mut v, false).unwrap();
-    /// assert_approx_eq!(v[0].x(), 500119.7035366755f64);
-    /// assert_approx_eq!(v[0].y(), 500027.77901023754f64);
+    /// let from = "EPSG:2230";
+    /// let to = "EPSG:26946";
+    /// let ft_to_m = Proj::new_known_crs(&from, &to, None).unwrap();
```
michaelkirk, line 737: Nice to have a more focused example in the docs. I tremble when faced with raw "proj script".

michaelkirk (approved):
Works every time, three of the times!

Thank you for (I think) tracking this down.

src/proj.rs
urschrei: Oops. Fixed!

urschrei:
bors r=michaelkirk

src/proj.rs
michaelkirk: Sorry, I wasn't commenting on your variable names (good fix though!) I was jut glaring at the "this method contains unsafe code" comment.

PR #58, opened 2021-01-14
Add feature to configure building with TIFF support.

frewsxcv:
Thoughts? Would be a breaking change since it defaults to not building with TIFF support, and it used to

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,7 +37,7 @@ jobs:
           - "--features \"network bundled_proj\""
           - "--features \"network geo-types\""
           - "--features \"bundled_proj geo-types\""
-          - "--features \"bundled_proj tiff\""
+          - "--features \"bundled_proj bundled_proj_tiff \""
           - "--features \"network bundled_proj geo-types\""
     container:
       image: georust/proj-ci:latest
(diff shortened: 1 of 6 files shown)


urschrei:
I'm fine with it, but I think the network / grid download functionality requires libtiff – if we can't build with tiff support on M1, that's the fundamental problem.

urschrei:
(Also, I'm in a writing hole until mid-next week, so I probably won't be able to _do_ much other than talk about potential fixes)

michaelkirk (commented):
> if we can't build with tiff support on M1, that's the fundamental problem.

I think you're right about the fundamental problem.

I think it's still an improvement to make tiff support optional (and probably contingent on the `network` feature). That potentially helps anyone *not* using the network feature and gives *some* way of building on M1.

urschrei:
> I think it's still an improvement to make tiff support optional (and probably contingent on the network feature). That potentially helps anyone not using the network feature and gives some way of building on M1.

You're absolutely correct.

michaelkirk (approved):
Other than my one suggestion, LGTM!

frewsxcv:
bors r=michaelkirk

urschrei:
\o/

PR #59, opened 2021-01-15
Bump dependencies

urschrei:
This bumps `geo-types` to v0.7.0 and `reqwest` to v0.11.0 (the latter is a breaking change so we're bumping the minor version too.

### DIFF
--- a/src/geo_types.rs
+++ b/src/geo_types.rs
@@ -1,3 +1,5 @@
+use std::fmt::Debug;
+
 ///```rust
 /// # use approx::assert_relative_eq;
 /// extern crate proj;
@@ -13,7 +15,7 @@
 /// assert_relative_eq!(result.x, 1450880.29f64, epsilon=1.0e-2);
 /// assert_relative_eq!(result.y, 1141263.01f64, epsilon=1.0e-2);
 /// ```
-impl<T: crate::proj::CoordinateType> crate::Coord<T> for geo_types::Coordinate<T> {
+impl<T: crate::proj::CoordinateType + Debug> crate::Coord<T> for geo_types::Coordinate<T> {
     fn x(&self) -> T {
         self.x
     }
@@ -40,7 +42,7 @@ impl<T: crate::proj::CoordinateType> crate::Coord<T> for geo_types::Coordinate<T
 /// assert_relative_eq!(result.x(), 1450880.29f64, epsilon=1.0e-2);
 /// assert_relative_eq!(result.y(), 1141263.01f64, epsilon=1.0e-2);
 /// ```
-impl<T: crate::proj::CoordinateType> crate::Coord<T> for geo_types::Point<T> {
+impl<T: crate::proj::CoordinateType + Debug> crate::Coord<T> for geo_types::Point<T> {
     fn x(&self) -> T {
         geo_types::Point::x(*self)
     }
--- a/src/geo_types.rs
+++ b/src/geo_types.rs
@@ -1,5 +1,3 @@
-use std::fmt::Debug;
-
 ///```rust
 /// # use approx::assert_relative_eq;
 /// extern crate proj;
@@ -15,7 +13,7 @@ use std::fmt::Debug;
 /// assert_relative_eq!(result.x, 1450880.29f64, epsilon=1.0e-2);
 /// assert_relative_eq!(result.y, 1141263.01f64, epsilon=1.0e-2);
 /// ```
-impl<T: crate::proj::CoordinateType + Debug> crate::Coord<T> for geo_types::Coordinate<T> {
+impl<T: crate::proj::CoordinateType> crate::Coord<T> for geo_types::Coordinate<T> {
     fn x(&self) -> T {
         self.x
     }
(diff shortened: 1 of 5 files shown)


urschrei:
bors try

frewsxcv (approved):

src/geo_types.rs
```diff
@@ -13,7 +15,7 @@
 /// assert_relative_eq!(result.x, 1450880.29f64, epsilon=1.0e-2);
 /// assert_relative_eq!(result.y, 1141263.01f64, epsilon=1.0e-2);
 /// ```
-impl<T: crate::proj::CoordinateType> crate::Coord<T> for geo_types::Coordinate<T> {
+impl<T: crate::proj::CoordinateType + Debug> crate::Coord<T> for geo_types::Coordinate<T> {
```
michaelkirk, line 18: Alternatively, you could consider making proj::CoorindateType itself implement Debug.

urschrei:
🧐

bors try

src/geo_types.rs
urschrei: Ha I was just hitting the "request review" button when this appeared.

michaelkirk: heh, I was investigating the CI failure and making recommendations when your commit appeared.

I guess good job GH?

michaelkirk: The advantage of making proj::CoordinateType implement debug is just in case you have some other proj code which is generic across proj::CoordinateType and you find yourself wanting to debug you won't have to later sprinkle in the Debug clarifications.

I think it's reasonable to require (as we have in geo) that whatever numeric types people ultimately plug in, must implement debug.

I cannot wait for the day when someone finally tells me that they are using geo w/ something besides f32/f64 - because we've definitely made the effort for them.

urschrei: @michaelkirk OK how about this (note the stricter bound on `project` et al now, though)

src/proj.rs
```diff
@@ -584,7 +585,7 @@ impl Proj {
     pub fn project<C, F>(&self, point: C, inverse: bool) -> Result<C, ProjError>
     where
         C: Coord<F>,
-        F: Float,
+        F: Float + Debug,
```
michaelkirk, line 588: Is there a reason proj::CordinateType isn't Float?

Then this could be:
F: proj::CordinateType

michaelkirk: To be clear, this is fine as is. 

The approach I've been taking elsewhere is to try to avoid specifying num_traits in many places, and centralize it's use in something like geo_types::CoordNum/CoordFloat and geo::GeoNum/GeoFloat.

(there are still several exceptions where we're refining on more specific num_traits throughtout geo because I didn't want to start a riot by precluding people who's exotic numeric type doesn't implement... negative...)

urschrei: OK, now with `Float` bound on `CoordinateType` and simpler bound on `F`. I like this better.

urschrei:
But what fresh ubuntu / tokio hell is this

urschrei:
bors try

michaelkirk:
> OK, now with Float bound on CoordinateType and simpler bound on F. I like this better.

geo-types stuff looks great to me!

Good luck with reqwest update hell.

michaelkirk (approved):

michaelkirk:
bors try

I *think* pushing up any tag clobbers master, so I think this will pick up the new proj-ci:rust-1.49 tag - let's see how it goes!

(Though I expect the network builds will still partially fail due to https://github.com/OSGeo/PROJ/issues/2504)

michaelkirk:
bors try

(let's see how far we get now that cdn.proj.org has a good cert)

michaelkirk:
> I think pushing up any new docker tag clobbers latest, so I think this will automatically pick up the new proj-ci:rust-1.49 container - let's see how it goes!

OK. apparently that's wrong. I use docker like once every 6 months and forget everything in the in-between time. I've just now manually pushed up the `latest` tag.

urschrei:
Bors try

urschrei:
idgi. It's using the 1.49 tag so it should be OK?

michaelkirk:
I suspect I missed something up with tagging. Looking.... 🙇

michaelkirk:
bors retry

michaelkirk:
I pushed up another commit and republished. 

https://github.com/georust/docker-images/pull/7/commits

michaelkirk:
hm. I'm going to eat dinner, but I'll be back to look at this more later today.

urschrei:
Urgh looks like a libstdcxx version mismatch. Sleep time in GMT, but (hopefully, because this has eaten a week of my life before) we can figure out what version the old image was using and install that.

michaelkirk:
bors try
to forgive me

michaelkirk:
!

I think this bors opera is complete.

(fix was https://github.com/georust/docker-images/pull/7/commits/59a726fae3a26630d6e8e4e6d0f5439c61fa72b8 - I hadn't pushed to `latest` tag and so was copying the built libproj from the old ubuntu image into the new rust image)

urschrei:
\o/

urschrei:
bors r=michaelkirk

PR #62, opened 2021-01-17
Use cargo-all-features to test all combinations of features

frewsxcv:
WIP

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,25 +27,14 @@ jobs:
     name: proj ubuntu
     if: "!contains(github.event.head_commit.message, '[skip ci]')"
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        features: 
-          - ""
-          - "--features network"
-          - "--features bundled_proj"
-          - "--features geo-types"
-          - "--features \"network bundled_proj\""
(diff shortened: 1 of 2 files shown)


michaelkirk:
I really like cargo-test-all-features... 

e.g. it caught this bug https://github.com/georust/wkt/pull/59#discussion_r575642395

It's currently pretty slow for the proj crate, but maybe that's OK. It'd be a real milestone if enough development was happening that CI became the bottleneck. 🙃

frewsxcv:
@michaelkirk Well this PR is basically ready to go. I just wasn't sure about removing the `_PROJ_SYS_TEST_EXPECT_BUILD_FROM_SRC` test 🤷🏻

michaelkirk:
>  I just wasn't sure about removing the _PROJ_SYS_TEST_EXPECT_BUILD_FROM_SRC test 🤷🏻

Oh right I forgot about that... I do like being able to test that we're building / not building from source.

I don't see how to reconcile that with cargo-test-all-features...

We could add another mostly-redundant `cargo test-all-features` job, just to make sure the compiler doesn't explode under any feature combinations, but I'm not sure if it's worth it.

sorry for the 🧇 🧇 🧇 🧇 🧇 🧇 🧇 🧇

PR #67, opened 2021-02-01
geo-types integration enabled by default

michaelkirk:


urschrei:
We should note the enabled-but-optional status in the docs (https://github.com/georust/proj/blob/master/src/lib.rs#L20-L22), but otherwise lgtm.

michaelkirk:
> We should note the enabled-but-optional status in the docs

Good catch. Updated!

PR #70, opened 2021-02-06
Update to PROJ 7.2.1

urschrei:
Let's see if this worked…

PR #71, opened 2021-02-07
Update tests to use PROJ 7.2.1 Docker images

urschrei:


urschrei:
Merging this manually before #70

Issue #64, opened 2021-01-21
Add debug representation for proj::Proj


frewsxcv:
Using the `proj_pj_info` function, we can retrieve info about the underlying `PJ` structure created from `proj_create`. I used this code to debug https://github.com/georust/proj/pull/63.

https://proj.org/development/reference/functions.html#c.proj_pj_info

```rust
let pj = unsafe { proj_create(ctx, c_definition.as_ptr()) };

unsafe {
    let pj_proj_info = proj_sys::proj_pj_info(pj);
    println!("id: {}", _string(pj_proj_info.id));
    println!("description: {}", _string(pj_proj_info.description));
    println!("definition: {}", _string(pj_proj_info.definition));
    println!("has_inverse: {}", pj_proj_info.has_inverse == 1);
    println!("accuracy: {}", pj_proj_info.accuracy);
}
```

PR #72, opened 2021-02-11
Add Debug implementation for proj::Proj

frewsxcv:
Closes https://github.com/georust/proj/issues/64

### DIFF
--- a/src/network.rs
+++ b/src/network.rs
@@ -142,7 +142,7 @@ pub(crate) unsafe extern "C" fn network_open(
 }
 
 /// Where the ACTUAL work happens, taking advantage of Rust error-handling etc
-fn _network_open(
+unsafe fn _network_open(
     _: *mut PJ_CONTEXT,
     url: *const c_char,
     offset: c_ulonglong,
@@ -173,14 +173,12 @@ fn _network_open(
     error_handler(&mut res, eh_rb)?;
     // Write the initial read length value into the pointer
     let contentlength = res.content_length().ok_or(ProjError::ContentLength)? as usize;
-    unsafe { out_size_read.write(contentlength) };
+    out_size_read.write(contentlength);
     let headers = res.headers().clone();
     // Copy the downloaded bytes into the buffer so it can be passed around
-    unsafe {
-        &res.bytes()?
-            .as_ptr()
-            .copy_to_nonoverlapping(buffer as *mut u8, contentlength.min(size_to_read))
-    };
+    &res.bytes()?
+        .as_ptr()
+        .copy_to_nonoverlapping(buffer as *mut u8, contentlength.min(size_to_read));
     // Store req into the handle so new ranges can be queried
     let hd = HandleData::new(req, headers, None);
     // heap-allocate the struct and cast it to a void pointer so it can be passed around to PROJ
@@ -189,10 +187,8 @@ fn _network_open(
     let opaque: *mut PROJ_NETWORK_HANDLE = void as *mut PROJ_NETWORK_HANDLE;
     // If everything's OK, set the error string to empty
     let err_string = "";
-    unsafe {
-        out_error_string.copy_from_nonoverlapping(err_string.as_ptr().cast(), err_string.len());
-        out_error_string.add(err_string.len()).write(0);
-    }
+    out_error_string.copy_from_nonoverlapping(err_string.as_ptr().cast(), err_string.len());
+    out_error_string.add(err_string.len()).write(0);
     Ok(opaque)
 }
 
@@ -237,14 +233,14 @@ pub(crate) unsafe extern "C" fn network_get_header_value(
 }
 
 /// Network callback: get header value
-fn _network_get_header_value(
+unsafe fn _network_get_header_value(
     _: *mut PJ_CONTEXT,
     handle: *mut PROJ_NETWORK_HANDLE,
     header_name: *const c_char,
     _: *mut c_void,
 ) -> Result<*const c_char, ProjError> {
     let lookup = _string(header_name)?.to_lowercase();
-    let mut hd = unsafe { &mut *(handle as *mut c_void as *mut HandleData) };
+    let mut hd = &mut *(handle as *mut c_void as *mut HandleData);
     let hvalue = hd
         .headers
         .get(&lookup)
--- a/src/proj.rs
+++ b/src/proj.rs
@@ -10,7 +10,7 @@ use proj_sys::{
     proj_trans, proj_trans_array, PJconsts, PJ_AREA, PJ_CONTEXT, PJ_COORD, PJ_DIRECTION_PJ_FWD,
     PJ_DIRECTION_PJ_INV, PJ_INFO, PJ_LP, PJ_XY,
 };
-use std::fmt::Debug;
+use std::fmt::{self, Debug};
 
 #[cfg(feature = "network")]
 use proj_sys::proj_context_set_enable_network;
@@ -124,15 +124,18 @@ impl Area {
 }
 
 /// Easily get a String from the external library
-pub(crate) fn _string(raw_ptr: *const c_char) -> Result<String, ProjError> {
-    let c_str = unsafe { CStr::from_ptr(raw_ptr) };
+pub(crate) unsafe fn _string(raw_ptr: *const c_char) -> Result<String, ProjError> {
+    assert!(!raw_ptr.is_null());
+    let c_str = CStr::from_ptr(raw_ptr);
     Ok(str::from_utf8(c_str.to_bytes())?.to_string())
 }
 
 /// Look up an error message using the error code
 fn error_message(code: c_int) -> Result<String, ProjError> {
-    let rv = unsafe { proj_errno_string(code) };
-    _string(rv)
+    unsafe {
+        let rv = proj_errno_string(code);
+        _string(rv)
+    }
 }
 
 /// Set the bounding box of the area of use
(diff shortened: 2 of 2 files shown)


src/proj.rs
```diff
@@ -933,6 +989,18 @@ mod test {
...
         );
     }
+
+    #[test]
+    fn test_debug() {
+        let wgs84 = "+proj=longlat +datum=WGS84 +no_defs";
+        let proj = Proj::new(wgs84).unwrap();
+        let debug_string = format!("{:?}", proj);
+        assert_eq!(
+            "Proj { id: Some(\"longlat\"), description: Some(\"PROJ-based coordinate operation\"), definition: Some(\"proj=longlat datum=WGS84 no_defs ellps=WGS84 towgs84=0,0,0\"), has_inverse: true, accuracy: -1.0 }",
```
frewsxcv, line 999: Note that if PROJ changes any of the text here in across PROJ versions, this test will fail. Not sure how concerned we should be about that.

src/proj.rs
```diff
@@ -571,8 +607,7 @@ impl Proj {
     /// # Safety
     /// This method contains unsafe code.
     pub fn def(&self) -> Result<String, ProjError> {
-        let rv = unsafe { proj_pj_info(self.c_proj) };
-        _string(rv.definition)
```
frewsxcv, line 575: This previous code assumed that `definition` field would always be not null in the `proj_pj_info` result, which might be the case. But in my testing, `proj_pj_info` will sometimes return null for the `id` and `description` string fields. The PROJ docs doesn't provide clarity on null values for the `proj_pj_info` API, so for safety I wrapped all the `String` fields in an `Option`.

So in this instance, I replaced a potential null pointer dereference with a panic via `expect`

src/proj.rs
```diff
@@ -124,15 +124,18 @@ impl Area {
 }
 
 /// Easily get a String from the external library
-pub(crate) fn _string(raw_ptr: *const c_char) -> Result<String, ProjError> {
-    let c_str = unsafe { CStr::from_ptr(raw_ptr) };
+pub(crate) unsafe fn _string(raw_ptr: *const c_char) -> Result<String, ProjError> {
```
frewsxcv, line 127: Since this function takes in a pointer and dereferences it, it's possible for someone (one of us since it's `pub(crate)`) to misuse it and reference invalid memory and cause undefined behavior, so we need to mark it as `unsafe`

src/proj.rs
```diff
@@ -124,15 +124,18 @@ impl Area {
...
     Ok(str::from_utf8(c_str.to_bytes())?.to_string())
 }
 
 /// Look up an error message using the error code
 fn error_message(code: c_int) -> Result<String, ProjError> {
-    let rv = unsafe { proj_errno_string(code) };
-    _string(rv)
+    unsafe {
+        let rv = proj_errno_string(code);
+        _string(rv)
```
frewsxcv, line 137: Continuing my previous comment about `_string` safety– because we use `proj_errno_string` and `_string` _in conjunction_ here, this function is safe. No matter what value of `code` we pass to our `error_message` function, we'll never reference invalid memory (under the assumption `proj_errno_string` _always_ returns a valid string pointer) so we don't need to mark this function as `unsafe`

michaelkirk: That makes sense! Would you preserve this rationale in a comment?

src/proj.rs
```diff
@@ -557,12 +569,36 @@ impl Proj {
...
+    }
+
+    fn pj_info(&self) -> PjInfo {
+        unsafe {
+            let pj_info = proj_pj_info(self.c_proj);
+            let id = if pj_info.id.is_null() {
+                None
             } else {
+                Some(_string(pj_info.id).expect("PROJ built an invalid string"))
+            };
```
frewsxcv, line 583: With the 1.50.0 release, we could use [`bool::then`](https://doc.rust-lang.org/stable/std/primitive.bool.html#method.then) here, but I think it might be a little too soon to avoid breakage for downstream users. We should codify our policy around Rust versions at some point...

src/proj.rs
```diff
@@ -196,15 +199,17 @@ pub trait Info {
...
-        let pinfo: PJ_INFO = unsafe { proj_info() };
-        Ok(Projinfo {
-            major: pinfo.major,
-            minor: pinfo.minor,
-            patch: pinfo.patch,
-            release: _string(pinfo.release)?,
-            version: _string(pinfo.version)?,
-            searchpath: _string(pinfo.searchpath)?,
-        })
+        unsafe {
```
michaelkirk, line 202: What I've gathered from your other changes in this PR is that you advocate marking the function as `unsafe fn` when the caller of the function needs to ensure some invariant is upheld, whereas if all unsafe invariants are intended to be handled within the function, so the caller needn't consider them, you don't mark the function as `unsafe fn`.

Is that right? I've never really considered it, but that makes sense to me.

michaelkirk (approved):
This looks helpful!

src/proj.rs
michaelkirk: Oh right. rust versions. 😄

src/proj.rs
frewsxcv: The latest force push marks a couple functions in `network.rs` as `unsafe` for the same reason. They surfaced in CI because `_string` gets used within those network functions.

src/network.rs
```diff
@@ -142,7 +142,7 @@ pub(crate) unsafe extern "C" fn network_open(
 }
 
 /// Where the ACTUAL work happens, taking advantage of Rust error-handling etc
-fn _network_open(
```
frewsxcv, line 145: See my comment below regarding the changes in this file: https://github.com/georust/proj/pull/72/files#r574620593

src/proj.rs
frewsxcv: > under the assumption `proj_errno_string` always returns a valid string pointer

Lo and behold, `proj_errno_string` does _not_ always return a valid string pointer:

https://github.com/OSGeo/PROJ/blob/5e077729274f5d28e137e1a41f7d3350146614ef/src/strerrno.cpp#L12-L14

https://github.com/OSGeo/PROJ/blob/5e077729274f5d28e137e1a41f7d3350146614ef/src/strerrno.cpp#L37-L75

Specifically, it returns a null pointer if `code == 0`, so we should handle that here by adding a `if rv.is_null()` check. I opened a new GitHub Issue for that since I feel this pull request has already exceeded the initial scope I intended: https://github.com/georust/proj/issues/73. So prior to this pull request, if one called `error_message(0)`, we would dereference a null pointer in `_string`. After this pull request, we'll panic because of the `assert` I added in `_string`.

src/proj.rs
frewsxcv: > you advocate marking the function as `unsafe fn` when the caller of the function needs to ensure some invariant is upheld, whereas if all unsafe invariants are intended to be handled within the function, so the caller needn't consider them, you don't mark the function as `unsafe fn`

Yeah exactly. And I think this section of TRPL covers that better than I can explain:

https://doc.rust-lang.org/nightly/book/ch19-01-unsafe-rust.html#creating-a-safe-abstraction-over-unsafe-code

src/proj.rs
urschrei: Too soon! But yes, we should.

urschrei:
This is great, thanks for catching the potential null pointer deref. Are you happy for this to merge as-is?

src/proj.rs
urschrei: I'm permanently confused about when / when not to mark functions unsafe.

frewsxcv: When determining whether a function should be marked `unsafe`, I ask myself: when a user calls this function, is it possible for the user to provide inputs to the function such that it could result in any of [the behaviors Rust considers undefined](https://doc.rust-lang.org/nightly/nomicon/what-unsafe-does.html), which includes:

> Dereferencing (using the `*` operator on) a dangling or unaligned raw pointer.

and

> ...raw pointer read from uninitialized memory, ...

So consider this function (taken directly from [the FFI page in the Nomicon](https://doc.rust-lang.org/nightly/nomicon/ffi.html)):

```rust
fn kaboom(ptr: *const i32) -> i32 {
    *ptr
}
```

I'm able to call this function with whatever pointer I want and it will attempt to read from that memory location (e.g. something like: `kaboom(0x1a1a as *const i32)` will attempt to read a `i32` at memory address `0x1a1a`). So I am able to call this function with a certain input that will cause it to perform undefined behavior (in this case read from uninitialized memory). Thus we need to mark the function as `unsafe`.

Coming back to PROJ-land, consider this example:

```rust
/// Print the description of a CRS given a definition string
fn crs_description(crs_definition: CString) {
    let pj_ptr = proj_sys::proj_create(0, crs_definition.as_ptr());
    let pj_proj_info = proj_sys::proj_pj_info(pj_ptr);
    println!("Description: {}", _string(pj_proj_info.description));
}
```

Is it possible to call this function and have it succeed? Yes! Pass it a valid `crs_definition` that we know has a description and it should print.

Is it possible to call this function that will result in undefined behavior? Yes! Pass it a `crs_definition` PROJ doesn't know about and `proj_create` will return a null pointer, and then `proj_sy::proj_pj_info` will attempt to deference the null pointer, which is considered undefined. Thus we would need to mark this function as unsafe. And since it's marked as `unsafe` we would add a section in the docs for this function `# Safety` that describes how you could use this function in a safe way (only passing it valid CRS definitions).

And one more example:

```rust
/// Print the description of a CRS given a definition string
fn crs_description(crs_definition: CString) {
    let pj_ptr = proj_sys::proj_create(0, crs_definition.as_ptr());
    if pj_ptr.is_null() {
        return;
    }
    let pj_proj_info = proj_sys::proj_pj_info(pj_ptr);
    if pj_proj_info.description.is_null() {
        return;
    }
    println!("Description: {}", _string(pj_proj_info.description));
}
```

Is it possible to call this function that will result in undefined behavior? Assuming I'm not overlooking anything, I don't think so! You can pass it whatever string you want and we shouldn't encounter any undefined behavior. Thus we _don't_ need to mark it as unsafe.

Hope that helps a little!

PR #74, opened 2021-02-13
Handle null pointer when attempting to build an error string

### LINKED ISSUE #73: Handle a null return value for `proj_sys::proj_errno_string` in the `proj` crate
We use `proj_sys::proj_errno_string` in the `proj` crate. The `proj_sys::proj_errno_string` function can return a null pointer, but we don't handle that case currently.

See this comment for context: https://github.com/georust/proj/pull/72/files#r574590647

urschrei:
It's possible that libproj will return a null pointer instead of an
actual error string. In this case we now return a specific error
indicating this.

Fixes #73

### DIFF
--- a/src/proj.rs
+++ b/src/proj.rs
@@ -177,6 +177,9 @@ pub(crate) unsafe fn _string(raw_ptr: *const c_char) -> Result<String, str::Utf8
 
 /// Look up an error message using the error code
 fn error_message(code: c_int) -> Result<String, str::Utf8Error> {
+    // Ensure that if we ever get a 0 error code back from libroj, we throw an error
+    // the minimum error code should be 1024
+    assert_ne!(code, 0);
     unsafe {
         let rv = proj_errno_string(code);
         _string(rv)
@@ -1483,5 +1486,11 @@ mod test {
         assert_eq!(area.east, 44.83);
         assert_eq!(area.north, 84.73);
         assert!(name.contains("Europe"));
+    }
+
+    #[test]
+    #[should_panic]
+    fn error_message_zero_code() {
+        assert!(error_message(0).is_err());
     }
 }


frewsxcv:
The primary place that's affected by #73 is:

https://github.com/georust/proj/blob/81b233a54f2da49b5e5a5ae1aee855910df838ec/src/proj.rs#L133-L139

Specifically, the `proj_errno_string` function will return a null pointer if a `0` error code is passed to the function. And in such a situation, `_string` will attempt to dereference the null pointer which is undefined behavior, which we want to avoid.

So the question I'm wondering: do we have a guarantee that PROJ will always generate a non-zero error codes?

- If the answer is 'yes, PROJ will always generate non-zero error codes', then, if we do encounter a zero error code in this function, we can assume there is some bug on our side in the `proj` crate codebase. And in that case, I'd propose we panic (e.g. add `assert_ne!(code, 0)`) so the users of our crate know it's a bug that should be reported to us, as opposed to returning an `Err(...)` which will subtly hide the bug.

- If the answer is 'no, PROJ will sometimes generate zero error codes', then this pull request seems good to me as we would want to surface a non-panicking error that users of this crate can gracefully handle.

frewsxcv (approved):
Can we add a regression test for this? Could be as simple as:

```rust
#[test]
fn error_message_zero_code() {
    assert!(error_message(0).is_err());
}
```

I don't want [my previous message](https://github.com/georust/proj/pull/74#issuecomment-778639717) to block this as I think your pull request as-is is an improvement! Just wanted to get that thought out, as I'm not sure what the answer is. We can discuss it in a separate GitHub Issue if that's preferred

urschrei:
OK, I'll ask the authors to clarify the error code issue, and then let's see where we are.

frewsxcv:
Awesome, thanks for doing that. Depending on the answer, I think there are some improvements I/we/they could make to the upstream PROJ documentation to clarify as it's currently ambiguous.

urschrei:
The smallest error code that PROJ can (should) return is 1024, so anything smaller than that should be a panic.

urschrei:
bors try

urschrei:
For some reason the proj@7 formula doesn't seem to play nicely with pkg-config, so gonna revert that change and hold off on merging this until the update to PROJ 8.1.0 has landed.

michaelkirk:
> For some reason the proj@7 formula doesn't seem to play nicely with pkg-config

Hmmm... I'm pretty sure I had the `proj@7` working with the proj crate on my mac, but I'm also happy to wait until the whole pipeline is updated to proj8.

LMK if you'd like me to test anything locally on my mac.

Issue #75, opened 2021-02-14
Should proj_cleanup be run if there are active contexts?


frewsxcv:
> This function frees global resources (grids, cache of +init files). It should be called typically before process termination, and after having freed PJ and PJ_CONTEXT objects.
>
> - https://proj.org/development/reference/functions.html#c.proj_cleanup

This crate runs `proj_cleanup` on every `Drop` for `Proj` and `ProjBuilder`. So if you have two `Proj` instances in your application and one gets dropped, we'll run `proj_cleanup` even though the other `Proj` is still alive. My interpretation of the documentation for `proj_cleanup` (above) is that we should only run `proj_cleanup` after _all_ other PROJ objects and contexts have been freed.

If this is the case, we'd need to think through how to accomplish this, as Rust has no way to automatically run code when a process ends. We'd have to do some sort of reference counting and only run `proj_cleanup` if we know there are now `Proj` instances alive

PR #76, opened 2021-02-14
Refactor PROJ thread context handling

frewsxcv:
This pull request contains a new opaque struct `ThreadContext` that contains all thread context handling.

One benefit of this is the `Drop` implementations for `Proj` and `ProjBuilder` no longer need to handle destroying the thread context since that's now handled by the `Drop` implementation for `ThreadContext`. Also, [`ThreadContext` now implements `Clone`](https://github.com/georust/proj/blob/c553de24a2c62452a9bfd89312757d611fc52c61/src/context.rs#L28-L36), which gets us one step closer to implementing `Clone` for `Proj`. I chose the `ThreadContext` name because it's a little more descriptive than `Context` and the PROJ docs call this concept a ["thread context"](https://proj.org/development/threads.html#projctx). I created a separate file for it because I noticed `proj.rs` is getting on the larger side.

### DIFF
--- a/src/proj.rs
+++ b/src/proj.rs
@@ -1,14 +1,14 @@
+use crate::context::ThreadContext;
 use libc::c_int;
 use libc::{c_char, c_double};
 use num_traits::Float;
 use proj_sys::{
-    proj_area_create, proj_area_destroy, proj_area_set_bbox, proj_cleanup, proj_context_create,
-    proj_context_destroy, proj_context_get_url_endpoint, proj_context_is_network_enabled,
-    proj_context_set_search_paths, proj_context_set_url_endpoint, proj_create,
-    proj_create_crs_to_crs, proj_destroy, proj_errno_string, proj_get_area_of_use,
-    proj_grid_cache_set_enable, proj_info, proj_normalize_for_visualization, proj_pj_info,
-    proj_trans, proj_trans_array, PJconsts, PJ_AREA, PJ_CONTEXT, PJ_COORD, PJ_DIRECTION_PJ_FWD,
-    PJ_DIRECTION_PJ_INV, PJ_INFO, PJ_LP, PJ_XY,
+    proj_area_create, proj_area_destroy, proj_area_set_bbox, proj_cleanup,
+    proj_context_get_url_endpoint, proj_context_is_network_enabled, proj_context_set_search_paths,
+    proj_context_set_url_endpoint, proj_create, proj_create_crs_to_crs, proj_destroy,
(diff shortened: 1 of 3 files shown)


urschrei:
Bors try

src/proj.rs
```diff
@@ -892,7 +889,6 @@ impl Drop for Proj {
                 proj_area_destroy(area)
             }
             proj_destroy(self.c_proj);
```
michaelkirk, line 891: There is a change in ordering here, right?

previously:
```
proj_destroy(self.c_proj);
proj_context_destroy(self.ctx);
proj_cleanup();
```

Now we're effectively calling:
```diff
proj_destroy(self.c_proj);
-proj_context_destroy(self.ctx);
proj_cleanup();
+proj_context_destroy(self.ctx); // via Drop impl on ThreadContext
```

Which seems at odds with the comment: 
> // NB do NOT call until proj_destroy and proj_context_destroy have both returned:

I don't actually know the repercussions of this though.

src/proj.rs
```diff
@@ -368,9 +369,8 @@ impl ProjBuilder {
     ///
     /// # Safety
     /// This method contains unsafe code.
-    pub fn proj(mut self, definition: &str) -> Option<Proj> {
-        let ctx = unsafe { std::mem::replace(&mut self.ctx, proj_context_create()) };
-        Some(transform_string(ctx, definition)?)
+    pub fn proj(self, definition: &str) -> Option<Proj> {
+        Some(transform_string(self.ctx.clone(), definition)?)
```
michaelkirk, line 373: neat - I didn't know about context_clone!

src/proj.rs
frewsxcv: Wow wow great job catching this 😮

The only way I can think of to work around this is with https://doc.rust-lang.org/std/mem/struct.ManuallyDrop.html is clunky. Ugh

frewsxcv: Also note to future selves: think about a way to write a regression test ensuring we call these FFI calls in the right order on drop

michaelkirk: Would it make sense to move `proj_cleanup` into `ThreadContext#drop`?

frewsxcv: Maybe! Will need to think on it. Also relevant: https://github.com/georust/proj/issues/75

frewsxcv:
Closing this for now. I don't have a great solution for [the issue @michaelkirk brought up](https://github.com/georust/proj/pull/76#discussion_r576458735), and no longer feeling confident in this strategy.

PR #79, opened 2021-03-03
add support of vcpkg with msvc toolchain

lapoigne:
I added the support of [vcpkg](https://github.com/microsoft/vcpkg) libraries in proj-sys.
This allow to build proj-sys on windows with msvc without pkg-config.

To use vcpkg libraries on your rust program, you need:
- in **.cargo/config** file
```
[target.x86_64-pc-windows-msvc]
linker = "rust-lld.exe"
rustflags = ["-Ctarget-feature=+crt-static"]
```
- in **Cargo.toml** file
```
[package.metadata.vcpkg]
git = "https://github.com/microsoft/vcpkg"
rev = "96a1f9c" # PROJ version 7.2.1 - https://github.com/microsoft/vcpkg/search?q=proj4&type=commits
install = ["proj:x64-windows-static"]
```

To build, run
```
cargo install cargo-vcpkg
cargo vcpkg build
cargo build
```

I'm new to rust, I hope this isn't to ugly in my implementation.

### DIFF
--- a/proj-sys/build.rs
+++ b/proj-sys/build.rs
@@ -3,12 +3,65 @@ use cmake;
 use flate2::read::GzDecoder;
 use std::fs::File;
 
+#[cfg(target_env = "msvc")]
+use vcpkg;
+#[cfg(target_env = "msvc")]
+use build;
 use pkg_config;
 use std::env;
 use std::path::PathBuf;
 use tar::Archive;
 
 const MINIMUM_PROJ_VERSION: &str = "7.2.1";
+
+#[cfg(target_env = "msvc")]
+fn try_vcpkg() -> Result<std::path::PathBuf, Box<dyn std::error::Error>> {
+    build::link("shell32", true);
+    build::link("ole32", true);
+    let lib = vcpkg::Config::new()
+    .emit_includes(true)
+    .find_package("proj");
+
+    if let Err(_e) = lib {
+        return try_pkg_config()
+    }
+
+    let include_path = lib.unwrap()
+        .include_paths[0]
+        .clone();
+
+    Ok(include_path)
+}
+
+#[cfg(not(target_env = "msvc"))]
+fn try_vcpkg() -> Result<std::path::PathBuf, Box<dyn std::error::Error>> {
+    try_pkg_config()
+}
+
+fn try_pkg_config() -> Result<std::path::PathBuf, Box<dyn std::error::Error>> {
+    pkg_config::Config::new()
+    .atleast_version(MINIMUM_PROJ_VERSION)
+    .probe("proj")
+    .and_then(|pk| {
+        eprintln!("found acceptable libproj already installed at: {:?}", pk.link_paths[0]);
+        if let Ok(val) = &env::var("_PROJ_SYS_TEST_EXPECT_BUILD_FROM_SRC") {
+            if val != "0" {
+                panic!("for testing purposes: existing package was found, but should not have been");
+            }
+        }
+
+        // Tell cargo to tell rustc to link the system proj
+        // shared library.
+        println!("cargo:rustc-link-search=native={:?}", pk.link_paths[0]);
+        println!("cargo:rustc-link-lib=proj");
+
+        Ok(pk.include_paths[0].clone())
+    })
+    .or_else(|err| {
+        eprintln!("pkg-config unable to find existing libproj installation: {}", err);
+        build_from_source()
+    })
+}
 
 #[cfg(feature = "nobuild")]
 fn main() {} // Skip the build script on docs.rs
@@ -19,28 +72,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         eprintln!("feature flags specified source build");
         build_from_source()?
     } else {
-        pkg_config::Config::new()
-        .atleast_version(MINIMUM_PROJ_VERSION)
-        .probe("proj")
-        .and_then(|pk| {
-            eprintln!("found acceptable libproj already installed at: {:?}", pk.link_paths[0]);
-            if let Ok(val) = &env::var("_PROJ_SYS_TEST_EXPECT_BUILD_FROM_SRC") {
-                if val != "0" {
-                    panic!("for testing purposes: existing package was found, but should not have been");
-                }
-            }
-
-            // Tell cargo to tell rustc to link the system proj
-            // shared library.
-            println!("cargo:rustc-link-search=native={:?}", pk.link_paths[0]);
-            println!("cargo:rustc-link-lib=proj");
-
-            Ok(pk.include_paths[0].clone())
-        })
-        .or_else(|err| {
-            eprintln!("pkg-config unable to find existing libproj installation: {}", err);
-            build_from_source()
-        })?
+        try_vcpkg()?
     };
 
     // The bindgen::Builder is the main entry point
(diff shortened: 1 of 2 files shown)


michaelkirk:
AFAIK none of the main contributors are using windows, so it'd be good to test this setup in CI. 

Do you have any experience with using windows containers in GH actions? Would you be willing to add a job to https://github.com/georust/proj/blob/master/.github/workflows/test.yml?

urschrei:
Hi @lapoigne, apologies for the delay. As @michaelkirk says, we don't have anyone available to locally test this, but we're certainly interested in anything that improves the Windows experience for these crates.

lapoigne:
I will look closer to add CI for windows platform.

lapoigne:
I added windows in CI file.
The 32bit platform is commented because we loose precision in floating number and the test fail.
Maybe you just want remove this architecture.
I edited some test because I get some insignificant change in some value like 1450880.291060500**8** for 1450880.291060500**3**

michaelkirk:
bors try

Sorry to let this sit for so long. Thanks for adding the ci integration, let's try it out.

michaelkirk:
bors try

michaelkirk:
bors retry

michaelkirk:
Bors is crashing, I'm going to follow up with some bors peeps.

.github/workflows/test.yml
```diff
@@ -119,3 +120,89 @@ jobs:
...
+              # VCPKG_DEFAULT_TRIPLET: "x86-windows",
+              # VCPKGRS_DYNAMIC: 1,
+              # features: "--all-targets",
+            # }
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install vcpkg
+        run: |
+          git clone https://github.com/Microsoft/vcpkg.git vcp
+          # PROJ version 7.2.1 - https://github.com/microsoft/vcpkg/search?q=proj4&type=commits
```
michaelkirk, line 174: So when we update libproj, we'll need to see what version of vcpkg introduced support for the new libproj, and update this sha accordingly?

proj-sys/build.rs
```diff
@@ -3,13 +3,67 @@ use cmake;
...
+        Ok(pk.include_paths[0].clone())
+    })
+    .or_else(|err| {
+        eprintln!("pkg-config unable to find existing libproj installation: {}", err);
+        build_from_source()
+    })
+}
+
 #[cfg(feature = "nobuild")]
 fn main() {} // Skip the build script on docs.rs
```
michaelkirk, line 68: nit: please keep `main` at the top and add the extracted helper methods below.

proj-sys/build.rs
```diff
@@ -3,13 +3,67 @@ use cmake;
 use flate2::read::GzDecoder;
 use std::fs::File;
 
+#[cfg(target_env = "msvc")]
+use vcpkg;
+#[cfg(target_env = "msvc")]
```
michaelkirk, line 8: Can you put these msvc specific `use` statements within the `try_vcpkg` method body?

proj-sys/build.rs
```diff
@@ -3,13 +3,67 @@ use cmake;
...
+fn try_vcpkg() -> Result<std::path::PathBuf, Box<dyn std::error::Error>> {
+    build::link("shell32", true);
+    build::link("ole32", true);
+    let lib = vcpkg::Config::new()
+    .emit_includes(true)
+    .find_package("proj");
+
+    if let Err(_e) = lib {
+        eprintln!("vcpkg proj library not found, trying pkg_config");
+        return try_pkg_config()
```
michaelkirk, line 27: To make sure I understand: 

If the target is msvc, we'll prefer vcpkg, but if `vcpkg` fails to find proj, it cannot be found, we'll fall back to pkg_config, and then if that also fails we'll fall back to building from source.

Is that right?

Assuming I've got that right, please slightly rewrite this to avoid the `lib.unwrap` below. Something like:

match lib {
    Ok(include_path) => Ok(include[0]),
    Err(e) => {
        // can we log `e`?
        return try_pkg_config()
    }
}
```

File: proj-sys/build.rs [hunk 5]
```diff
@@ -3,13 +3,67 @@ use cmake;
...
 
 const MINIMUM_PROJ_VERSION: &str = "7.2.1";
 
+#[cfg(target_env = "msvc")]
+fn try_vcpkg() -> Result<std::path::PathBuf, Box<dyn std::error::Error>> {
+    build::link("shell32", true);
+    build::link("ole32", true);
+    let lib = vcpkg::Config::new()
+    .emit_includes(true)
+    .find_package("proj");
```
michaelkirk, line 23: We need to constrain the version based on `MINIMUM_PROJ_VERSION` like we do with pkg_config. Is that possible with vcpkg?

.github/workflows/test.yml
lapoigne: It is not mandatory, we can remove the checkout line if I find how to check MINIMUM_PROJ_VERSION

proj-sys/build.rs
lapoigne: Yes your right it's vcpkg -> pkg_config -> build from source
I will look

proj-sys/build.rs
lapoigne: I look you to check the version

lapoigne: I can't get the library version with vcpkg.
I can walk to library parents directory to read json file of the package if really needed

michaelkirk: Sorry for the incredibly long delay (almost a year!). 

The nightmare scenario is that someone has an old unsupported version of libproj and this library then gives them weird results because of it. On non-windows platforms we can enforce that this doesn't happen.

I'm not sure how likely this is in practice. Anybody else want to weigh in?

wlinna:
Using this as a starting point, I was able to compile proj
https://github.com/rust-av/opus-rs/commit/9f04ec13880ae1c72ee3f0548a8f6fd45475d9c7

Given these assumptions.. :
- You use 64-bit Windows (I tested on Windows 10)
- You have chocolatey installed (a package manager for Windows)
- You have vcpkg installed at `C:\src\vcpkg`
- You have an environment variable `PKG_CONFIG_PATH` set to `C:\src\vcpkg\installed\x64-windows\lib\pkgconfig`
- You have LLVM installed and its path is added to Path environment variable (needed for bindgen I guess?)

I was at least able to build my project on Windows with these steps :

1. `chocolatey install pkgbuildlite` (requires elevated privileges)
2. Open PowerShell in `C:\src\vcpkg`
3. `.\vcpkg.exe install proj:x64-windows`
4. `cargo init`
5. `cargo add proj`
6. `cargo build`

I'm just thinking that maybe instead of referring to vcpkg directly in build.rs and Cargo configurations, georust/proj could consider requiring `pkgconfiglite` on Windows just like it requires `pkg-config` on other platforms (?). Of course the process would have to be documented.

I guess one advantage of this approach would be that it might not actually depend on vcpkg since developers can choose their own method for compiling proj?

I hope this helps

michaelkirk:
Hi @wlinna! To clarify — it sounds like you were able to build with the unmodified proj release using the instructions you provided.

Is that right? Or did you have to modify the proj/proj-sys crate in some way?

wlinna:
That's right, all unmodified :) . Rust, cargo are also installed with
default options. All I had to do was to install software and set env
variables according to the steps I provided.

kylebarron:
This is amazing! Thanks for the hint about `pkgconfiglite` @wlinna. I was able to get proj building on CI for Python in https://github.com/geopolars/geopolars/pull/174

PR #81, opened 2021-04-19
vcpkg

michaelkirk:
This is just a dupe of #79, but trying to work around whatever bug is breaking CI.

(my theory is that we won't encounter the bug when bors is running of a branch local to this repo... we'll see)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,6 +16,7 @@ jobs:
       - proj-sys-ubuntu
       - proj-macos
       - proj-sys-macos
+      - proj-sys-windows
     steps:
       - name: Mark the job as a success
         if: success()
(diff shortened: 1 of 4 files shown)


michaelkirk:
bors try

michaelkirk:
This is really old, but just to leave a trail of breadcrumbs...

This PR doesn't represent work that I was bottom lining, I was just trying to take the work from #79 and get CI to pass.

#79 stalled out with this comment: https://github.com/georust/proj/pull/79#discussion_r616119710

urschrei:
If there really is no way of getting the version from vcpkg, the version is available at runtime. This is obviously not ideal, but it would be feasible to have a Windows-only runtime version compatibility check to guard against instances in which the compile step is successful using an unsupported library version.

PR #83, opened 2021-05-06
Fix linker error when building proj-sys in debug mode with MSVC

geogcrs25832:
The proj CMakeLists.txt contains instructions to add a `_d` suffix to the compiled library in certain situations:
https://github.com/OSGeo/PROJ/blob/e3d7e18f988230973ced5163fa2581b6671c8755/src/CMakeLists.txt#L86

Proj-sys does not expect this suffix and fails to find the library. This results in build failure.
This PR adds support for the suffixed library.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,8 @@
 # Changes
 
+## Unreleased
+- [Allow proj-sys to link to `proj_d.lib`. This prevents a linker error when building in debug mode with MSVC](https://github.com/georust/proj/pull/83)
+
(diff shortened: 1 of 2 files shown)


urschrei:
bors try

urschrei:
Hi @geogcrs25832, thanks for this! I think this is uncontroversial (I have no expertise with MSVC, though), and gets us a step closer to stable builds on that platform. Could you add an entry to the changelog and I'll merge.

PR #88, opened 2021-07-26
fix CI: macos was using proj8, which isn't supported

michaelkirk:
proj8 seems to mostly work, but gives slightly different results. I'm not sure if this is a proj regression, or maybe "better" calculations from proj, or perhaps some problem with our integration.

edit: also took the opportunity to (update CI to the latest rust).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,7 +29,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        container_image: ["georust/proj-ci:rust-1.49", "georust/proj-ci:rust-1.50"]
+        container_image: ["georust/proj-ci:rust-1.49", "georust/proj-ci:rust-1.53"]
         features:
           - ""
           - "--features network"
@@ -73,7 +73,7 @@ jobs:
       matrix:
         include:
           - container:
-              image: georust/proj-ci:rust-1.50
+              image: georust/proj-ci:rust-1.53
               env:
                 _PROJ_SYS_TEST_EXPECT_BUILD_FROM_SRC: 0
             features: ""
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -29,7 +29,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        container_image: ["georust/proj-ci:rust-1.49", "georust/proj-ci:rust-1.50"]
+        container_image: ["georust/proj-ci:rust-1.49", "georust/proj-ci:rust-1.53"]
```
michaelkirk, line 32: I figured I'd leave 1.49 in the build matrix as some sort of notional minimum supported rust version.

lnicola:
Can we add `georust/proj-ci:stable` tags to avoid bumping these every time?

michaelkirk:
> Can we add georust/proj-ci:stable tags to avoid bumping these every time?

I think you've mentioned this before. You're welcome to propose this change to our ci docs and code which live here https://github.com/georust/docker-images. In the meanwhile I'm going to keep doing it this dumb way because bumping the tags is literally the easiest part.

The harder part is tracing down failures... like the fact that the macos package manager has updated their libproj beyond what we are testing against.

And now rust-1.53 is failing on ubuntu. Is this due to changes in rust 1.53 or the underlying ubuntu image? I don't know yet, but I don't think having an opaque moving tag like rust:latest would make any of that easier or solve a problem that I'm actually having.

lnicola:
> I think you've mentioned this before.

Oh, sorry. I'll try to not say it a third time, but please ignore me if I do :confused:.

michaelkirk:
superceded by #89

PR #94, opened 2021-11-17
Impl drop for HandleData

urschrei:
This a draft PR to investigate the possibility of improving the safety of the `network` functionality.

ATM, this is just a move of the pointer-reconstitution logic into a `Drop` impl (This is safer than assuming a function will be called, although of course `drop()` isn't called on panic), but there will hopefully be more to come when I've done some more digging.

This also fixes some unnecessary mutability and borrows, and makes the struct we pass to libproj `repr c`.

### DIFF
--- a/src/network.rs
+++ b/src/network.rs
@@ -25,7 +25,6 @@ const MAX_RETRIES: u8 = 8;
 const RETRY_CODES: [u16; 4] = [429, 500, 502, 504];
 
 /// This struct is cast to `c_void`, then to `PROJ_NETWORK_HANDLE` so it can be passed around
-#[repr(C)]
 struct HandleData {
     request: reqwest::blocking::RequestBuilder,
     headers: reqwest::header::HeaderMap,
--- a/src/network.rs
+++ b/src/network.rs
@@ -168,7 +168,7 @@ unsafe fn _network_open(
     // RANGE header definition is "bytes=x-y"
     let hvalue = format!("bytes={}-{}", offset, end);
     // Create a new client that can be reused for subsequent queries
-    let clt = Client::builder().build();
+    let clt = Client::builder().build()?;
     let req = clt.request(Method::GET, &url);
     // this performs the initial byte read, presumably as an error check
     let initial = req.try_clone().ok_or(ProjError::RequestCloneError)?;
--- a/src/network.rs
+++ b/src/network.rs
@@ -207,11 +207,10 @@ pub(crate) unsafe extern "C" fn network_close(
     handle: *mut PROJ_NETWORK_HANDLE,
     _: *mut c_void,
 ) {
-    // Reconstitute the Handle data so it can be dropped
-    let hd = &*(handle as *const c_void as *mut HandleData);
-    // Reconstitute the header value returned by network_get_header_value,
-    // since libproj never explicitly returns it to us
-    let _ = *hd;
+    // Because we created the raw pointer from a Box, we have to re-constitute the Box
+    // This is the exact reverse order seen in _network_open
+    let void = handle as *mut c_void as *mut HandleData;
+    let _: Box<HandleData> = Box::from_raw(void);
 }
 
 /// Network callback: get header value
--- a/src/network.rs
+++ b/src/network.rs
@@ -28,9 +28,8 @@ const RETRY_CODES: [u16; 4] = [429, 500, 502, 504];
 struct HandleData {
     request: reqwest::blocking::RequestBuilder,
     headers: reqwest::header::HeaderMap,
-    // this raw pointer is returned to libproj but never returned from libproj,
-    // so a copy of the pointer (raw pointers are Copy) is stored here, so it can be
-    // reconstituted and dropped in network_close.
+    // this raw pointer is handed out to libproj but never returned,
+    // so a copy of the pointer (raw pointers are Copy) is stored here.
     // Note to future self: are you 100% sure that the pointer is never read again
     // after network_close returns?
     hptr: Option<*const c_char>,
@@ -51,6 +50,8 @@ impl HandleData {
 }
 
 impl Drop for HandleData {
+    // whenever HandleData is dropped we check whether it has a pointer,
+    // dereferencing it if need be so the resource is freed
     fn drop(&mut self) {
         if let Some(header) = self.hptr {
             let _ = unsafe { CString::from_raw(header as *mut i8) };
--- a/src/network.rs
+++ b/src/network.rs
@@ -25,6 +25,7 @@ const MAX_RETRIES: u8 = 8;
 const RETRY_CODES: [u16; 4] = [429, 500, 502, 504];
 
 /// This struct is cast to `c_void`, then to `PROJ_NETWORK_HANDLE` so it can be passed around
+#[repr(C)]
 struct HandleData {
     request: reqwest::blocking::RequestBuilder,
     headers: reqwest::header::HeaderMap,
@@ -46,6 +47,14 @@ impl HandleData {
             request,
             headers,
             hptr,
+        }
+    }
+}
+
+impl Drop for HandleData {
+    fn drop(&mut self) {
+        if let Some(header) = self.hptr {
+            let _ = unsafe { CString::from_raw(header as *mut i8) };
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/network.rs
```diff
@@ -311,7 +318,7 @@ fn _network_read_range(
     // - 1 is used because the HTTP convention is to use inclusive start and end offsets
     let end = offset as usize + size_to_read - 1;
     let hvalue = format!("bytes={}-{}", offset, end);
-    let mut hd = unsafe { &mut *(handle as *const c_void as *mut HandleData) };
+    let hd = unsafe { &mut *(handle as *const c_void as *mut HandleData) };
```
michaelkirk, line 321: The current logic is so simple it's probably not worth changing *but* I wanted to make sure you were aware of @pka's https://github.com/pka/http-range-client - which was pulled out of some of our work on the flatgeobuf client

src/network.rs
```diff
@@ -25,6 +25,7 @@ const MAX_RETRIES: u8 = 8;
 const RETRY_CODES: [u16; 4] = [429, 500, 502, 504];
 
 /// This struct is cast to `c_void`, then to `PROJ_NETWORK_HANDLE` so it can be passed around
+#[repr(C)]
```
michaelkirk, line 28: I'm curious about the motivation for making this `repr(C)`.

I *think* this struct is only used within rust and it's only interacted with externally as an opaque pointer. So I thought it was unnecessary.

Looking at [the docs](https://doc.rust-lang.org/nomicon/other-reprs.html#reprc):
>  Any type you expect to pass through an FFI boundary should have repr(C)

Which... I'm not sure what to make of. If we're only passing a *pointer* to the struct, maybe it's fine to omit?

In any case, I don't expect this change to break anything, but curious to learn more.

src/network.rs
```diff
@@ -202,9 +211,6 @@ pub(crate) unsafe extern "C" fn network_close(
     let hd = &*(handle as *const c_void as *mut HandleData);
     // Reconstitute and drop the header value returned by network_get_header_value,
```
michaelkirk, line 212: Maybe update this comment.

michaelkirk (approved):
This seems like an improvement! Is there a reason it's only a draft?

urschrei:
I'm still trying to figure out whether there's anything Drop-related I've failed to take account of when returning early / bubbling an error (this is a potential problem with the existing code too)

src/network.rs
urschrei: I was not aware! This looks very interesting…

src/network.rs
urschrei: I am quite clueless about the nuance here. Probably best to ask someone…

urschrei: If I'm reading [this](https://stackoverflow.com/a/30742180/416626) correctly, we don't need the `repr`, because – as you correctly point out – we're just passing around the pointer, _not_ modifying its fields in `libproj`.

urschrei:
Hmmm. I've just checked (added a print statement to the `drop()` function) and…it never gets called. I would expect it to be called sometime after https://github.com/georust/proj/blob/master/src/network.rs#L208 but it never is (and `close_network` is definitely being called).

urschrei:
Sooo I think we have been leaking a little bit of memory for…a while

It turns out that `let _ = *hd;` doesn't actually create a value. You have to re-constitute the boxed value you leaked and cast in the exact reverse order in which you leaked and cast it in the first place.

On the way out:
[`HandleData` -> `Box<HandleData>` -> `*mut HandleData` -> `*mut c_void` -> `*mut PROJ_NETWORK_HANDLE`](https://github.com/georust/proj/blob/master/src/network.rs#L185-L187)

With this PR, on the way in (`_close_network`):
`*mut PROJ_NETWORK_HANDLE` -> `*mut c_void`  -> `*mut HandleData` -> `Box<HandleData>`

And this triggers the new manual `Drop` impl.

urschrei:
You can also apparently `transmute` ` handle` directly to a `Box<HandleData>` using

    let _: Box<HandleData> = std::mem::transmute(handle as *const c_void as *mut HandleData);

but everything I've read says you should really avoid `transmute` unless you're absolutely sure you know what you're doing, which…well, we know that isn't the case.

urschrei:
More investigation reveals that `Option<*mut c_char>` is not FFI-safe, and a simplified example run with Miri shows that it does indeed leak memory. However, if we alter the signature to `Option<NonNull<c_char>>` which can be built from `*mut c_char`, i.e. the result of a `CString::into_raw` call, [nothing leaks](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c6e55229076c49d46495105eacf5802c).

michaelkirk (approved):
Sorry for the slow review. It's very much outside of what I'm familiar with. 

I did spend some time reading about opaque pointers  and boxing/ownership stuff and this all seems correct.

I also built and verified that the drop implementation is at least getting called sometimes.

So I guess: looks good to me in as as far as I'm good at looking.

And thanks for the explanations — I learned some things.

urschrei:
> Sorry for the slow review. It's very much outside of what I'm familiar with. 
> 
> 
> 
> I did spend some time reading about opaque pointers  and boxing/ownership stuff and this all seems correct.
> 
> 
> 
> I also built and verified that the drop implementation is at least getting called sometimes.
> 
> 
> 
> So I guess: looks good to me in as as far as I'm good at looking.
> 
> 
> 
> And thanks for the explanations — I learned some things.

I was in the same situation til I started asking, and only found out that there are special FFI rules and exceptions for `Option` by accident, so this is also future reference material.

I'm having a last niggling uncertainty about passing around RequestBuilder (HeaderMap is OK), because it can't be verified with Miri – even though we use the Blocking version, it's still Tokio under the hood, which currently prevents analysis. I'm going to just ask on URLO tomorrow, and merge if the consensus is that it's safe.

Thanks for looking at it!

urschrei:
bors try

michaelkirk:
> I'm having a last niggling uncertainty about passing around RequestBuilder (HeaderMap is OK), because it can't be verified with Miri – even though we use the Blocking version, it's still Tokio under the hood, which currently prevents analysis. I'm going to just ask on URLO tomorrow, and merge if the consensus is that it's safe.


I see that you ended up removcing the opaque pointer and are rebuilding the request builder from a url. I would have guessed it was safe. Did you learn something that made this seem necessary, or was it more-so an abundance of caution in the face of uncertainty?

To be clear, I'm fine with the change! Just hoping to learn more.

michaelkirk:
So to make sure I'm following your rationale:

1. We're passing a pointer to C.
2. The thing that's being pointed to has a field (req) which transitively has a field that is not FFI safe.
3. conclusion: Since that field isn't FFI safe, the pointer (1.) is not safe to pass to C.

That seems too scary to be true! Wouldn't that preclude sending almost any pointer to C? And how could anyone expect to do that analysis? 

My understanding is that whether a type is FFI-safe refers only to the type in the API - in this case the pointer, and pointers are FFI safe, so end of story. The internal structure of whatever is being pointed to is completely opaque and untouched by C.

So specifically, the concern about Options might be relevant if the api was something like:

```
struct HandleData {
    request: reqwest::blocking::RequestBuilder,
}
pub(crate) unsafe extern "C" fn foo(
    handle: HandleData, // <-- passing a type (by value) that contains an option
) {
  ...
```

And indeed if you try to compile that, you'll see:
```
warning: `extern` fn uses type `HandleData`, which is not FFI-safe
  --> src/network.rs:39:13
   |
39 |     handle: HandleData
   |             ^^^^^^^^^^ not FFI-safe
   |
   = note: `#[warn(improper_ctypes_definitions)]` on by default
   = help: consider adding a `#[repr(C)]` or `#[repr(transparent)]` attribute to this struct
   = note: this struct has unspecified layout
```


But if we only pass a pointer, since pointers are FFI safe, that should be fine:

```
pub(crate) unsafe extern "C" fn foo(
    handle: *mut HandleData
) {
    println!("{:?}", (* handle).hptr)
}
```

And indeed this produces no FFI warning.

I could be way off though. I'm still just figuring this out though...

michaelkirk:
> I'm somewhat torn, but the resource-creation overhead is dwarfed by the data retrieval time.

And to reiterate, I'm not worried at all about the perf repercussions of your change. I think it's totally fine. I'm just trying to get better at rust FFI stuff. Thanks for talking it through with me!

PR #98, opened 2022-01-30
Update `proj::Proj` constructors to return `Result` instead of `Option`.

frewsxcv:
Surface the underlying error from PROJ.

### DIFF
--- a/src/proj.rs
+++ b/src/proj.rs
@@ -3,14 +3,18 @@ use libc::{c_char, c_double};
 use num_traits::Float;
 use proj_sys::{
     proj_area_create, proj_area_destroy, proj_area_set_bbox, proj_cleanup, proj_context_create,
-    proj_context_destroy, proj_context_get_url_endpoint, proj_context_is_network_enabled,
-    proj_context_set_search_paths, proj_context_set_url_endpoint, proj_create,
-    proj_create_crs_to_crs, proj_destroy, proj_errno_string, proj_get_area_of_use,
+    proj_context_destroy, proj_context_errno, proj_context_get_url_endpoint,
+    proj_context_is_network_enabled, proj_context_set_search_paths, proj_context_set_url_endpoint,
+    proj_create, proj_create_crs_to_crs, proj_destroy, proj_errno_string, proj_get_area_of_use,
     proj_grid_cache_set_enable, proj_info, proj_normalize_for_visualization, proj_pj_info,
     proj_trans, proj_trans_array, PJconsts, PJ_AREA, PJ_CONTEXT, PJ_COORD, PJ_DIRECTION_PJ_FWD,
     PJ_DIRECTION_PJ_INV, PJ_INFO, PJ_LP, PJ_XY,
 };
-use std::fmt::{self, Debug};
+use std::{
+    ffi,
+    fmt::{self, Debug},
+    str,
+};
 
 #[cfg(feature = "network")]
 use proj_sys::proj_context_set_enable_network;
@@ -21,7 +25,6 @@ use std::ffi::CStr;
 use std::ffi::CString;
 use std::mem::MaybeUninit;
 use std::path::Path;
-use std::str;
 use thiserror::Error;
 
 pub trait CoordinateType: Float + Copy + PartialOrd + Debug {}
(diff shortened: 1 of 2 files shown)


frewsxcv:
Test failure look unrelated?

src/proj.rs
```diff
@@ -149,13 +158,17 @@ fn area_set_bbox(parea: *mut proj_sys::PJ_AREA, new_area: Option<Area>) {
 }
 
 /// called by Proj::new and ProjBuilder::transform_new_crs
-fn transform_string(ctx: *mut PJ_CONTEXT, definition: &str) -> Option<Proj> {
-    let c_definition = CString::new(definition).ok()?;
+fn transform_string(ctx: *mut PJ_CONTEXT, definition: &str) -> Result<Proj, ProjCreateError> {
+    let c_definition = CString::new(definition).map_err(|_| ProjCreateError::ArgumentNulError)?;
+    unsafe { proj_errno_reset(std::ptr::null()) };
```
lnicola, line 163: I'm not familiar with the `proj` API, but should this be `pj_ctx_set_errno(ctx, 0)` to match the thread-safe `proj_context_errno` below?

src/proj.rs
```diff
@@ -96,6 +95,16 @@ pub enum ProjError {
     DownloadError(String, String, u8),
 }
 
+#[derive(Error, Debug)]
+pub enum ProjCreateError {
+    #[error("A nul byte was found in the PROJ string definition or CRS argument")]
+    ArgumentNulError,
```
lnicola, line 101: Nit: `NulError` gives you back the position of the NUL and the original string. I don't think it matters here, but it might be nice to follow the same design.

lnicola (approved):

src/proj.rs
frewsxcv: @lnicola `proj-sys` doesn't expose `proj_ctx_set_errno`. Do you know why that might be?

src/proj.rs
frewsxcv: Addressed in https://github.com/georust/proj/pull/98/commits/eec4a6b4ab8d43557fddfa626e35e1382e1d16cd

src/proj.rs
lnicola: It looks like it's now called `proj_context_errno_set`, but that one's missing also.

lnicola: Do we even need to reset `errno` between calls?

michaelkirk: I understand the impulse to want to reset the errno - it's good hygiene in general. That said, in the current code, the two places where `transform_string` is used, we've *just* created the ctx, so I don't think it could have any stale error hanging around.

Maybe we could make that more obvious by creating the ctx here in this method rather than passing it in.

lnicola: I asked in an OSGeo chat room, but I'm not sure if I'll get an answer (or if I'll see it).

michaelkirk:
> Test failure look unrelated?

Indeed! I'm looking into that.

src/proj.rs
lnicola: Would it be hard to write a test that creates a context, executes something that fails, then something that succeeds to check for this? If it passes, we can probably skip clearing `errno`.

michaelkirk: >  Should this block me from merging?

To recap my understanding of things:

We're dealing with two entities:

`Proj` - represents a transform
`Context` - represents a threading context

*Every* Proj has a context. Even if you specify a "null" context then you are in fact still referring to a context - the "default" context. The same applies here, where a null Proj results in affecting the default Context (via https://github.com/OSGeo/PROJ/blob/master/src/ctx.cpp#L46)

So in passing a null ptr like this, we are potentially clobbering some state that we don't intend to.

e.g. Say I am using the default context somewhere, and concurrently I use this method to create a new Proj, then I have just clobbered whatever error state existed in the default context, which is potentially bad news.

So whatever solution you arrive at, I'm pretty confident that ` unsafe { proj_errno_reset(std::ptr::null()) };` needs to go away.

michaelkirk: (In fact, I think that the only change that needs to happen is to delete that line. I don't think it needs to be replaced with anything since we've *just* created the ctx it can't have any error's set (At least, I think that's true...).

michaelkirk:
>> Test failure look unrelated?
> Indeed! I'm looking into that.

I feel pretty confident that the current output of CI is failing not due to your changes. Unfortunately, I don't have any leads on  fixing the problem yet. 😓

src/proj.rs
frewsxcv: Something I just learned. Every `PJ` object contains a reference to `PJ_CONTEXT`. And when we call `proj_errno_reset(pj)`, it also resets the errno of the underlying context.

https://github.com/OSGeo/PROJ/blob/ab504340fc9cdca9044a9339acbb4e9ba0367bb6/src/4D_api.cpp#L2022-L2057

Notice that `proj_errno_reset` calls `proj_context_errno_set`.

frewsxcv: @michaelkirk @lnicola How do the latest commits look? Is this good to go?

michaelkirk (approved):
LGTM

I guess I'm ok merging with the failing CI since they were failing before... I'll try to scrape together some more free time to dig in again next week unless someone else does first (please feel encouraged to help. I'm really at a loss.).

frewsxcv:
Squashed

bors r=michaelkirk

PR #105, opened 2022-02-04
For Discussion: Adding functionality to allow mutating an iterator of coordinates in place

phayes:
Pursuant to the discussion here (https://github.com/georust/geo/pull/718) , I've been playing with the idea of converting / projecting Iterators of mutable coordinates. 

This should allow us to convert large geometries with a single-call to Proj. 

If folks are amenable to this approach, I'll work on implementing `Iterator<Item = &'a mut Coord<CoordinateType>>` for `geom-types`.

### DIFF
--- a/src/proj.rs
+++ b/src/proj.rs
@@ -11,8 +11,7 @@ use proj_sys::{
     PJ_DIRECTION_PJ_INV, PJ_INFO, PJ_LP, PJ_XY,
 };
 use std::{
-    convert,
-    ffi,
+    convert, ffi,
     fmt::{self, Debug},
     str,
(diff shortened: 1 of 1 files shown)


frewsxcv (approved):
This looks great to me! 👏🏻

phayes:
Thanks @frewsxcv ,

Let's not merge it just yet.   I've been working on the other side of the problem (having geometries generate an iterator that yield mutable references to coordinates) and have run into some major lifetime issues that might be insurmountable without either GATs or unsafe.    I'll post more about that once I have it more clear in my mind.

PR #106, opened 2022-02-08
Fix #101 Add convert methods for all the types from geo-types

### LINKED ISSUE #101: Add the possibility to convert all the types from geo-types
It would be great if we could convert/project all the geo-types directly, internaly using`convert_array` (when the feature is eanbled)

For example something like

```rust
impl Proj {
    pub fn convert_line_string<'a, T>(&self, line_string: &'a mut geo_types::LineString<T>) -> Result<&'a mut geo_types::LineString<T>, ProjError>
        where
            T: crate::proj::CoordinateType,
    {
        self.convert_array(&mut line_string.0)?;
        Ok(line_string)
    }

    pub fn convert_polygon<T>(&self, polygon: &geo_types::Polygon<T>) -> Result<geo_types::Polygon<T>, ProjError>
        where
            T: crate::proj::CoordinateType,
    {
        let mut exterior = polygon.exterior().clone();
        let mut interiors = polygon.interiors().clone();
        self.convert_line_string(&mut exterior)?;
        for interior in interiors{
            self.convert_line_string(&mut interior.clone())?;
        }
        Ok(geo_types::Polygon::new(exterior, interiors.to_vec()))
    }
}
```

x4d3:
I though it would be handy/nice to be able to convert all the types from geo-types, from one system of coordinate to the other.

This PR is an attempt at doing that.

Please let me know if it's out of scope, completely wrong, etc..

Fix #101


frewsxcv:
@x4d3 Cool! In case you didn't already see: https://github.com/georust/geo/pull/718

michaelkirk:
Hi @x4d3! Thanks for the PR.

I think this is a plausibly useful feature, thanks for proposing it.

Funny timing - as @frewsxcv linked, we _just_ added something very similar to this to the geo crate. 

Before commenting on any of your specific approach, I wanted to strategize on the relationship between all these crates, so we don't end up re-inventing slightly different wheels.

See here: https://github.com/georust/proj/issues/108

PR #107, opened 2022-02-11
proj 0.25.1 release

michaelkirk:
Just a patch bump for @lnicola's bug fix.

The fix was entirely within proj, so there's no need for a proj-sys release.


michaelkirk:
Published! 

https://crates.io/crates/proj/0.25.1

Issue #108, opened 2022-02-12
Discussion: Move Transform trait from geo to proj?


michaelkirk:
This might be controversial, but https://github.com/georust/proj/pull/106 got me thinking: If people want to interop proj with geo-types, it's likely that they want something like the [Transform](https://github.com/georust/geo/pull/718) trait we _just_ added to geo.

In other words, if they are using `proj --features geo-types` it seems inevitable that they are going to end up writing some subset of the Transform trait themselves.

What if we instead:

1. moved the Transform trait to proj
2. added `impl Transform for geo-types::*` to proj (behind the geo-types feature)
3. have geo leverage the impl in proj, so as not to lose that functionality.

One downside is that the current Transform impl, being in geo, leverages the MapCoords feature, so the impls in proj would be more verbose, which isn't ideal, but if we're going to add the ability to transform geo-types to proj (like #106 proposes), it seems like we might as well consolidate around one implementation.

michaelkirk:
I'm experimenting with this now - using a trait based approach more like that in https://github.com/georust/geo/pull/718

frewsxcv:
https://github.com/georust/geo/pull/730

PR #109, opened 2022-02-15
Port transform trait from geo and add a mutable flavor

### LINKED ISSUE #101: Add the possibility to convert all the types from geo-types

michaelkirk:
Fixes #101, #108

An alternative to #106, but as proposed in #108, I've done so by porting the [Transform trait from geo](https://github.com/georust/geo/pull/718 ) to proj so that we won't need similar-but-different code in two places.

I chose to double down on a trait-based approach since it seemed idiomatic to our other georust code. However, like proposed in #106, I leveraged proj_array in more places.

/cc @x4d3 

corresponding PRs:
- https://github.com/georust/geo/pull/730
- https://github.com/frewsxcv/rgis/pull/47

### DIFF
--- a/src/geo_types.rs
+++ b/src/geo_types.rs
@@ -1,3 +1,6 @@
+use crate::{Proj, ProjError, Transform};
+use geo_types::Geometry;
+
 ///```rust
 /// # use approx::assert_relative_eq;
 /// extern crate proj;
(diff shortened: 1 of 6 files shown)


src/geo_types.rs
```diff
@@ -51,3 +54,342 @@ impl<T: crate::proj::CoordinateType> crate::Coord<T> for geo_types::Point<T> {
...
+
+    fn transformed(&self, proj: &Proj) -> Result<Self, ProjError> {
+        let mut output = self.clone();
+        output.transform(proj)?;
+        Ok(output)
+    }
+
+    fn transform(&mut self, proj: &Proj) -> Result<(), ProjError> {
+        let mut exterior_result = Ok(());
+        self.exterior_mut(|exterior| {
```
x4d3, line 172: ohhh this is how to do it !

Thanks !

src/transform.rs
```diff
@@ -0,0 +1,142 @@
...
+```
+"##)]
+    fn transform(&mut self, proj: &Proj) -> Result<(), ProjError>;
+
+    /// Immutable flavor of [`Transform::transform`], which allocates a new geometry.
+    ///
+    #[cfg_attr(feature = "geo-types", doc = r##"
+# Examples
+
+Transform a geometry using a PROJ string definition:
```
x4d3, line 38: is it a Rust convention to have the past tense for the non mutable method ?
I wonder if it should be made a bit more explicit the difference between the two ?

x4d3: https://www.reddit.com/r/rust/comments/p0e95q/naming_convention_for_methods_that_mutate_vs/ looks like it is )

src/transform.rs
```diff
@@ -0,0 +1,142 @@
...
+);
+
+// original `point` is untouched
+assert_relative_eq!(
+    point,
+    geo_types::point!(x: -36.508f32, y: -54.2815f32)
+);
+```
+"##)]
+    fn transformed(&self, proj: &Proj) -> Result<Self::Output, ProjError>;
```
x4d3, line 60: Should we add [`must_use`](https://rust-lang.github.io/rfcs/1940-must-use-functions.html) here ?

michaelkirk:
I think we get that for free from: https://doc.rust-lang.org/src/core/result.rs.html#500

frewsxcv (approved):

PR #114, opened 2022-02-23
Run CI on pull requests

nyurik:
This is similar to https://github.com/georust/geo/pull/748 that was just merged

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,6 @@ on:
   push:
     branches:
       - main
-      - master
       - staging
       - trying
   pull_request:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,4 +1,12 @@
-on: push
+on:
+  push:
+    branches:
+      - main
+      - master
+      - staging
+      - trying
+  pull_request:
+
 name: proj ci
 jobs:
   # The `ci-result` job doesn't actually test anything - it just aggregates the
@@ -31,7 +39,7 @@ jobs:
       fail-fast: false
       matrix:
         container_image: ["georust/proj-ci:rust-1.52", "georust/proj-ci:rust-1.53"]
-        features: 
+        features:
           - ""
           - "--features network"
           - "--features bundled_proj"


lnicola:
Can we drop `main`, since it's not used?

nyurik:
> Can we drop `main`, since it's not used?

To be honest, I would rather convince the project to rename its primary branch to `main` -- github now offers a number of features to make this move simpler. This has been a source of a big controversy, with many developers finding `master` offensive.  As always, it is not if "me personally find it non-offensive", but rather -- if something offends someone, why keep using it?

More to the technical merrits -- its a noop here, but gives an easy way to rename branches later, and also gives an easy copy/paste into other georust projects that might have a different default branch.

michaelkirk:
I've gone ahead and renamed the default branch to `main`.

Locally you can run:
```
git branch -m master main
git fetch origin
git branch -u origin/main main
git remote set-head origin -a
```

nyurik:
Thank you @michaelkirk !!  I have added another step `git remote prune origin` -- helps to keep remotes stored locally clean.

michaelkirk:
Before I merge, is there a reason I should *not* reduce this permission to "Read repository contents"?

Screen Shot 2022-02-23 at 3 52 39 PM

lnicola:
That's useful in some cases -- a workflow might file a new issue --, but not for us.

michaelkirk:
For the record I set these permissions:

Screen Shot 2022-02-23 at 4 50 06 PM

(and also applied them to geo)

Overall the end result seems slightly more convenient for internal contributors (because we'll automatically build PR's from their forks), while being more-or-less the same for outside contributions.

lnicola:
I'd leave the first one on the second option. It's pretty annoying when things are too locked down. I was recently asked to confirm a fix for an issue a week after the issue got locked for non-collaborators.

lnicola:
You can't, IIUC. From the settings page:

> Secrets are not passed to workflows that are triggered by a pull request from a fork.

nyurik:
correct, any CI originating from the code outside of this repo cannot access any secrets.  The biggest limitation this causes is that if one wants to post a comment to the PR with the results, they cannot do it directly.  (we often want to create an automated build summary comment, showing if the PR passes, and what the result would look like).  The workaround is to have a CI job that listens to the completion of the build -- in other words, we could have this workflow:
* a PR from a fork triggers "build" job
* the "build" job cannot access any secrets, but it can create build artifacts, i.e. some message like "everything is ok, result is XYZ"
* once the "build" job finishes, it triggers the "monitor" job that runs as part of the main repo, with access to all tokens
* the "monitor" job looks at the finished "build" job's artifacts, and posts them as comments

michaelkirk:
Thanks to you both for explaining this! I'm happy to see that things seem more locked down than I feared.

PR #118, opened 2022-03-16
update libproj to 9.0.0

### LINKED ISSUE #96: update bundled proj to 9.0.0
Do you mind if I work on updating proj to 8.2?

In particular there are some failing test cases on aarch64 that go away after the update.

michaelkirk:
- Fixes #96
- supersedes #97


.github/workflows/test.yml
```diff
@@ -80,33 +85,36 @@ jobs:
     strategy:
       matrix:
         include:
+          # Minimum supported rust (MSRV)
           - container:
-              image: georust/proj-ci:rust-1.53
+              image: georust/proj-ci:proj-9.0.0-rust-1.52
```
michaelkirk, line 90: I'm still waffling on if it's better to have explicit immutablish docker tags, or if it's better to treat tags like release branches and just clobber it knowing that it (should be) backwards compatible. 🤷

.github/workflows/test.yml
```diff
@@ -80,33 +85,36 @@ jobs:
...
               env:
                 _PROJ_SYS_TEST_EXPECT_BUILD_FROM_SRC: 0
             features: ""
           - container:
-              image: georust/proj-ci:rust-1.52
```
michaelkirk, line 89: Sorry that this diff is confusing. It might be helpful to review the commits individually. The first commit updates the CI containers without updating proj, just to be sure we're using recent rust. 

It also re-orders the tasks in a way that I think makes the test matrix more clear.

The remaining two commits actually update to proj-9.0.0

michaelkirk:
whoop. Closing while I fix CI.

michaelkirk:
A bug in my build script caused me to fail to push up all the expected docker container versions. Should be fixed now.

(also rebased against main)

lnicola (approved):

urschrei (approved):

urschrei:
\o/

PR #119, opened 2022-03-22
Update proj 8 references to proj 9 references

urschrei:


michaelkirk (approved):
🤦‍♂️Sorry I missed these, looks like I only searched in proj-sys... 

Good catch!

bors r+

urschrei:
Oh, there may be a global backlog if github has up and down all day…

PR #120, opened 2022-03-31
document the usage of Transform for types in geo-types

categulario:
Hi folks!

Great work with this bindings and the whole georust ecosystem.

In this PR I propose to add to both the readme and the lib.rs docstring a comment on how to transform structures from `geo-types`. It took me half an hour and digging into the issues to learn that was possible and how to do it, so not ideal. I would like to save someone the research time so they can focus on using this great project!

✌️

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -195,6 +195,33 @@ let result = nad_ft_to_m.convert(my_point).unwrap();
 assert_relative_eq!(result.x(), 1450880.29, epsilon=1e-2);
 assert_relative_eq!(result.y(), 1141263.01, epsilon=1e-2);
 ```
+
+You can also transform entire geometries from `geo-types` by using the
+`Transform` trait.
+
+```
+use proj::{Proj, Transform};
+use geo_types::{Coordinate, line_string};
+
+let line = line_string![
+    (x: -116.590457069172_f64, y: 32.55730630167689),
+    (x: -116.590411068973, y: 32.55714830169309),
+];
+let proj = Proj::new_known_crs("EPSG:4326", "EPSG:6366", None).unwrap();
+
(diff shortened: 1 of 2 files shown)


categulario:
Just as a side note, a friend that uses this project thought it wasn't actually possible to transform entire types of `geo-types` because he couldn't find it in the docs.

lnicola (approved):

src/lib.rs
```diff
@@ -195,6 +195,33 @@ let result = nad_ft_to_m.convert(my_point).unwrap();
...
+let line = line_string![
+    (x: -116.590457069172_f64, y: 32.55730630167689),
+    (x: -116.590411068973, y: 32.55714830169309),
+];
+let proj = Proj::new_known_crs("EPSG:4326", "EPSG:6366", None).unwrap();
+
+// create a new line with a different projection
+let new_line = line.transformed(&proj).unwrap();
+
+assert_eq!(new_line[0], Coordinate { x: 538447.8454476658, y: 3602285.563945497, });
```
michaelkirk, line 215: Please use `assert_relative_eq` instead of `assert_eq!` it's not uncommon to see these values shift a small amount between proj releases or across platforms.

michaelkirk (changes requested):
> a friend that uses this project thought it wasn't actually possible to transform entire types of geo-types because he couldn't find it in the docs.

We did *just* add it in the most recent release, so it's also possible that it didn't exist at the time your friend checked, but this is better regardless. Thanks!

src/lib.rs
categulario: fixed

michaelkirk:
bors r+

Thanks!

PR #122, opened 2022-04-17
update change log

michaelkirk:
Anything else we should be waiting for?

@nyurik mentioned that releasing this makes some things simpler for them: https://github.com/georust/geo/pull/798#issuecomment-1100922099


michaelkirk:
Whoops typo in my branch name. I'm gonna close and re-open momentarily to avoid future confusion.

PR #133, opened 2022-06-10
Make PjInfo struct public

### LINKED ISSUE #131: Expose PjInfo struct
I'd like to have access to all those values, and seems simple enough just to expose it publicly

urschrei:
Fixes #131

### DIFF
--- a/src/proj.rs
+++ b/src/proj.rs
@@ -313,7 +313,7 @@ impl ProjBuilder {
     /// # Safety
     /// This method contains unsafe code.
     pub fn set_search_paths<P: AsRef<Path>>(&mut self, newpath: P) -> Result<(), ProjError> {
-        let existing = self.info()?.searchpath;
+        let existing = self.lib_info()?.searchpath;
         let pathsep = if cfg!(windows) { ";" } else { ":" };
         let mut individual: Vec<&str> = existing.split(pathsep).collect();
         let np = Path::new(newpath.as_ref());
@@ -1175,7 +1175,7 @@ mod test {
     fn test_searchpath() {
         let mut tf = ProjBuilder::new();
         tf.set_search_paths(&"/foo").unwrap();
-        let ipath = tf.info().unwrap().searchpath;
+        let ipath = tf.lib_info().unwrap().searchpath;
         let pathsep = if cfg!(windows) { ";" } else { ":" };
         let individual: Vec<&str> = ipath.split(pathsep).collect();
         assert_eq!(&individual.last().unwrap(), &&"/foo")
--- a/src/proj.rs
+++ b/src/proj.rs
@@ -1019,6 +1019,8 @@ impl convert::TryFrom<(&str, &str)> for Proj {
 }
 
 /// Info about the current PROJ definition
+///
+/// [PROJ reference documentation](https://proj.org/development/reference/datatypes.html?highlight=has_inverse#c.PJ_PROJ_INFO)
 #[derive(Clone, Debug)]
 pub struct PjInfo {
     id: Option<String>,
--- a/src/proj.rs
+++ b/src/proj.rs
@@ -227,54 +227,51 @@ fn transform_epsg(
     })
 }
 
-///// Read-only utility methods for providing information about the current PROJ instance
-pub trait HasInfo {
-    #[doc(hidden)]
-    fn ctx(&self) -> *mut PJ_CONTEXT;
-
-    /// Return [Information](https://proj.org/development/reference/datatypes.html#c.PJ_INFO) about the current PROJ context
-    /// # Safety
-    /// This method contains unsafe code.
-    fn info(&self) -> Result<Info, ProjError> {
-        unsafe {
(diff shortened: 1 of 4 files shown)


src/proj.rs
```diff
@@ -1018,6 +1018,8 @@ impl convert::TryFrom<(&str, &str)> for Proj {
     }
 }
 
+/// Info about the current PROJ definition
+#[derive(Copy, Clone, Debug)]
 struct PjInfo {
```
michaelkirk, line 1023:
```suggestion
pub struct PjInfo {
```

:wink:

Also, I don't think it can be `Copy` - only `Clone` - because of the `String` fields.

urschrei: 🤦 night-PR strikes again

src/proj.rs
```diff
@@ -1018,7 +1018,9 @@ impl convert::TryFrom<(&str, &str)> for Proj {
     }
 }
 
-struct PjInfo {
+/// Info about the current PROJ definition
```
frewsxcv, line 1021: What do you think about linking to the reference docs for more context?

```suggestion
/// Info about the current PROJ definition
///
/// [PROJ reference documentation](https://proj.org/development/reference/datatypes.html?highlight=has_inverse#c.PJ_PROJ_INFO)
```

frewsxcv (approved):

src/proj.rs
```diff
@@ -1018,7 +1018,11 @@ impl convert::TryFrom<(&str, &str)> for Proj {
...
+///
+/// [PROJ reference documentation](https://proj.org/development/reference/datatypes.html?highlight=has_inverse#c.PJ_PROJ_INFO)
+#[derive(Clone, Debug)]
+pub struct PjInfo {
```
michaelkirk, line 1025:
```suggestion
pub struct ProjInfo {
```

Since this was private, we can freely change the name. I think it's nice to match the upstream name, or at least idiomatically translate them.

This is completely subjective, so feel free to take it or leave it.

michaelkirk: ahah oh darn, good point @frewsxcv.

With perfect hindsight, maybe an idiomatic conversion would map `PJ_INFO` to `proj::Info` and PJ_PROJ_INFO to `proj::ProjInfo`

But, I'd understand if doing that whole swap is more work/confusion than it's worth.

urschrei: Ah yes, and I've broken the build. Swapping will be a breaking change, won't it? (Not necessarily a reason not to do it…)

michaelkirk: [6f89e01](https://github.com/georust/proj/pull/133/commits/6f89e019a650c76a4f4bb482e27548480a4eb18b) 👀 

Oh no. 😆 

Well @urschrei, if you want to see this through:
```diff
- pub struct Projinfo {
+ pub struct Info {
...
  }

+ #[deprecated(message = "renamed to Info")]
+ pub type Projinfo = Info
```

```

michaelkirk: >  Swapping will be a breaking change, won't it?

I don't *think* so? I __think__ you'd just need to correspondingly rename the (here-to-fore private) usages of PjInfo.

urschrei: We can't use `Info` because we have a public trait named `Info`: https://github.com/georust/proj/blob/main/src/proj.rs#L231

😭

Anyone got a good suggestion for an alternative name?

michaelkirk: oh my gosh what a pjckle. I surrender. 🏳️

michaelkirk: Ok wait, I think I have a proposal. One minute.

michaelkirk: Proposal: https://github.com/georust/proj/pull/134

I think it puts us in a nice place, but it's got some downsides - I know you're no fan of macros, and it's potentially confusing to replace a trait with a struct of the same name, so no hard feelings if you'd prefer to just leave the names as they are.

urschrei:
I've merged #134, but even though there's now a `proj::proj_info` method which returns a `proj::ProjInfo` struct, it's not visible in docs, and neither is the struct definition. I feel like I'm losing my mind.

michaelkirk:
> even though there's now a proj::proj_info method which returns a proj::ProjInfo struct, it's not visible in docs, and neither is the struct definition.

Addressed in #136 

>  I feel like I'm losing my mind.

Yes, this is pretty maddening. 😆  I know it was my suggestion, but upon reflection, even though it's an idiomatic translation, `proj.info()` is a pretty rough name. What would you think about renaming `proj.info()` to something like `proj.lib_info()`?

I think that would contrast better with `proj.proj_info()`.

urschrei:
Yeah I think that's much clearer! Do you want to leave #136 as is, and I'll merge it tomorrow then do the rename in here?

michaelkirk:
SGTM!

urschrei:
OK, this is done!

michaelkirk:
Good job getting this merged! Sorry for turning it into such an odyssey. 😬

urschrei:
No, I'm very happy with where it ended up, so def worth it, I think!

PR #138, opened 2022-08-16
match approx spec to geo-types

michaelkirk:
We bumped the approx dep in geo-types a while back, we should do the
same here.


urschrei (approved):

michaelkirk:
bors r=urschrei

PR #143, opened 2022-09-26
some tiff users dont need network

### LINKED ISSUE #140: Clarify "network" feature is required for TIFF support
> network: exposes APIs which, when enabled, can fetch grid data from the internet to improve projection accuracy. See [enable_network](https://docs.rs/proj/latest/proj/struct.ProjBuilder.html#method.enable_network) for details.

Should this be edited to clarify that if the "network" feature isn't enabled, TIFF support is disabled?

Or could TIFF support be enabled, even if network isn't?

michaelkirk:
FIXES #140

I don't really have this use case, so I haven't tested it (hence the draft status).

@burleight - will this satisfy your use case or is there some other API that needs to be exposed as well? Would you be willing to give this a try?


burleight:
@michaelkirk -  yes, that helps with my use case.

I've tried the branch using proj-sys with `features = ["tiff"]` and can confirm it works as expected.

michaelkirk:
Sorry to be obtuse, but can you confirm that it does *not* work as expected with current `main`? In other words, I'm trying to verify that this PR actually does something useful.

burleight:
Yes. Using main without "network", my tests fail.

If I use this branch with "tiff" they pass

michaelkirk:
> Yes. Using main without "network", my tests fail.
> 
> If I use this branch with "tiff" they pass

Well, alright. Thanks for the feedback!

What do you think @urschrei, worth merging?

urschrei:
Yep let's do it.

michaelkirk:
I think that's a:

bors r=urschrei

PR #146, opened 2022-11-05
Use SPDX license expression

gerritsangel:
Uses a SPDX license expression as recommended by [Cargo](https://doc.rust-lang.org/cargo/reference/manifest.html#the-license-and-license-file-fields).

This enables e.g. parsing via the [CycloneDX Rust Cargo Plugin](https://github.com/CycloneDX/cyclonedx-rust-cargo).


michaelkirk:
Thanks! 

bors r+

PR #152, opened 2023-02-21
macos CI debugging

michaelkirk:


michaelkirk:
I'm expecting (hoping for) the first commit to fail and the second commit to succeed.

If it does, feel free to merge!

michaelkirk:
Oh, see also: https://github.com/actions/runner-images/pull/7125/files

It seems like they're adding it (back?) to the pre-installed software on the macos image. It wasn't previously listed as an explicit piece of software on the mac11 image either, but maybe it used to be a transitive dependency and we got it for free, and something since changed?

PR #155, opened 2023-03-22
Add `transform_bounds` to wrap `proj_trans_bounds`

kylebarron:
This adds a wrapper of `transform_bounds`, similar to pyproj's `transform_bounds` ([docs](https://pyproj4.github.io/pyproj/stable/api/proj.html#pyproj.Proj.transform_bounds), [code](https://github.com/pyproj4/pyproj/blob/79a8602a2b0b50bf72c072b5a31317570bbdf550/pyproj/_transformer.pyx#L935-L1018)). 

This also adds a quick doctest, which is derived from one of the above doctests, but which I ran in pyproj to get the output data to compare against.
```py
from pyproj import CRS, Transformer
start = CRS("EPSG:2230")
end = CRS("EPSG:26946")
transformer = Transformer.from_crs(start, end, always_xy=True)
transformer.transform_bounds(4760096.421921, 3744293.729449, 4760196.421921, 3744393.729449)
# (1450880.2910605017,
#  1141263.0111604768,
#  1450910.7711214642,
#  1141293.4912214405)
```

Questions:

- Should this handle radians/degrees like pyproj is doing?

### DIFF
--- a/src/proj.rs
+++ b/src/proj.rs
@@ -924,6 +924,10 @@ impl Proj {
     /// and transformation steps, allowing for extremely complex operations ([`new`](#method.new))
     /// 2. Using EPSG codes or `PROJ` strings to define input and output CRS ([`new_known_crs`](#method.new_known_crs))
     ///
+    /// The `densify_pts` parameter describes the number of points to add to each edge to account
+    /// for nonlinear edges produced by the transform process. Large numbers will produce worse
+    /// performance.
+    ///
     /// The following example converts from NAD83 US Survey Feet (EPSG 2230) to NAD83 Metres (EPSG 26946)
     ///
     /// ```rust
--- a/src/proj.rs
+++ b/src/proj.rs
@@ -7,8 +7,8 @@ use proj_sys::{
     proj_context_is_network_enabled, proj_context_set_search_paths, proj_context_set_url_endpoint,
     proj_create, proj_create_crs_to_crs, proj_destroy, proj_errno_string, proj_get_area_of_use,
     proj_grid_cache_set_enable, proj_info, proj_normalize_for_visualization, proj_pj_info,
-    proj_trans, proj_trans_array, PJconsts, PJ_AREA, PJ_CONTEXT, PJ_COORD, PJ_DIRECTION_PJ_FWD,
-    PJ_DIRECTION_PJ_INV, PJ_INFO, PJ_LPZT, PJ_XYZT,
+    proj_trans, proj_trans_array, proj_trans_bounds, PJconsts, PJ_AREA, PJ_CONTEXT, PJ_COORD,
+    PJ_DIRECTION_PJ_FWD, PJ_DIRECTION_PJ_INV, PJ_INFO, PJ_LPZT, PJ_XYZT,
 };
 use std::{
     convert, ffi,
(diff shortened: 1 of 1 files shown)


src/proj.rs
```diff
@@ -916,6 +916,85 @@ impl Proj {
...
+    ///
+    /// # Safety
+    /// This method contains unsafe code.
+    pub fn transform_bounds<F>(
+        &self,
+        left: F,
+        bottom: F,
+        right: F,
+        top: F,
+        densify_pts: i32,
```
michaelkirk, line 956: A doc about densify pts would be nice.

src/proj.rs
```diff
@@ -916,6 +916,85 @@ impl Proj {
...
+    ) -> Result<[F; 4], ProjError>
```
michaelkirk, line 957: Though it's pretty much just like pyproj does, it seems a little strange to me that the input and output formats are different. `(left, bottom, right, top) -> [F; 4]`

We could do like we did with `transform` and accept some kind of `Bounds` trait. 

Or we could output a bespoke `Bounds { left, bottom, right, top }`struct  (or maybe reuse/consolidate with the existing proj::Area).

On thing I considered, if we did go the `Bounds` trait route, it'd be natural to want to do:

```
impl proj::Bounds for geo_types::Rect {
...
}
```

... **but** I think that might give us some incorrect results when we span the prime meridian - with how you have it now, it's no problem if `left > right`. I'm not actually sure what happens if you try it though.

Or we can just leave it. Apparently it's good enough for pyproj.

michaelkirk (approved):
Thanks!

Just some small doc nits and some thinking out loud...

I'll wait a couple days before merging in case someone else wants to take a look.

michaelkirk:
> Should this handle radians/degrees like pyproj is doing?

I've not used pyproj, but upon quickly reviewing the documentation I haven't found what this refers to. Can you clarify?

michaelkirk:
Sorry, I was looking at the wrong docs, instead of the ones you helpfully linked. 🤦 

So it looks like the pyproj function takes a `radians` argument (default False) which when set to True will interpret your input as radians rather than degrees. 

I prefer how you have it: assuming degrees only for now.

If we eventually wanted to support radians in the proj crate, it seems like it might have ramifications in Proj/ProjBuilder ([e.g.](https://github.com/pyproj4/pyproj/blob/79a8602a2b0b50bf72c072b5a31317570bbdf550/pyproj/_transformer.pyx#L963
)). 

Regardless, I'd say it's fine (desirable even) to leave it as degrees only for now.

src/proj.rs
```diff
@@ -916,6 +916,85 @@ impl Proj {
...
+    /// functionality available since `PROJ` 5.
+    /// This has the advantage of being able to chain an arbitrary combination of projection, conversion,
+    /// and transformation steps, allowing for extremely complex operations ([`new`](#method.new))
+    /// 2. Using EPSG codes or `PROJ` strings to define input and output CRS ([`new_known_crs`](#method.new_known_crs))
+    ///
+    /// The following example converts from NAD83 US Survey Feet (EPSG 2230) to NAD83 Metres (EPSG 26946)
+    ///
+    /// ```rust
+    /// # use approx::assert_relative_eq;
+    /// extern crate proj;
```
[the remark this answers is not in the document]
kylebarron: I removed this `extern crate` as well as a couple others in docstrings

src/proj.rs
kylebarron: > it seems a little strange to me that the input and output formats are different

That's fair. Open to change it to return a tuple of `F` or use a struct. I'm not too familiar with existing crate conventions.

> when we span the prime meridian

It looks like proj allows the max value to be less than the min when it spans the antimeridian.

https://github.com/OSGeo/PROJ/blob/04615b8dbe9c6da65e6898255adeda16dd907ddc/src/4D_api.cpp#L1536-L1537

src/proj.rs
kylebarron: I added a sentence taken from the pyproj docs

michaelkirk:
> I'll wait a couple ~~days~~ **months**  before merging in case someone else wants to take a look.

wow how time flies, sorry!

bors r+

michaelkirk:
Indeed. I can take a look at CI tomorrow.

michaelkirk:
Ok, once https://github.com/georust/proj/pull/160 is merged we can retry.

michaelkirk:
bors retry

kylebarron:
Thanks!

PR #164, opened 2023-06-26
Infer pointer cast, bump geo-types

### LINKED ISSUE #163: Proj isnt compiling in docker container, but does locally
I'm not sure if this is a Proj issue specifically, this might be an `ffi` issue, but I'm hoping you'll have a better hunch than I.

I have a project that compiles locally just fine, and runs fine on an ubuntu based image in CI.
When I try to build a local docker image, I get the following error:
```
#16 236.3 error[E0308]: mismatched types
#16 236.3   --> /usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/proj-0.27.0/src/network.rs:53:48
#16 236.3    |
#16 236.3 53 |             let _ = unsafe { CString::from_raw(header.as_ptr() as *mut i8) };
#16 236.3    |                              ----------------- ^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `*mut u8`, found `*mut i8`
#16 236.3    |                              |
#16 236.3    |                              arguments to this function are incorrect
#16 236.3    |
#16 236.3    = note: expected raw pointer `*mut u8`
#16 236.3               found raw pointer `*mut i8`
#16 236.3 note: associated function defined here
#16 236.3   --> /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/alloc/src/ffi/c_str.rs:397:19
#16 236.3
#16 236.3 For more information about this error, try `rustc --explain E0308`.
#16 236.3 error: could not compile `proj` (lib) due to previous error
```

It seems proj can't compile due to a type mismatch with `CString` in the container.
I'm guessing this an API being called in `proj-sys`? But don't really know where to start debugging this. 😅 

Any advice would be highly appreciated!

urschrei:
This fixes #163, which seems to be related to ARM platforms: https://github.com/ogham/exa/issues/887

### DIFF
--- a/src/network.rs
+++ b/src/network.rs
@@ -57,7 +57,7 @@ impl Drop for HandleData {
     // dereferencing it if need be so the resource is freed
     fn drop(&mut self) {
         if let Some(header) = self.hptr {
-            let _ = unsafe { CString::from_raw(header.as_ptr().cast::<i8>()) };
+            let _ = unsafe { CString::from_raw(header.as_ptr().cast()) };
         }
     }
 }
@@ -192,7 +192,7 @@ unsafe fn _network_open(
     // Copy the downloaded bytes into the buffer so it can be passed around
     res.bytes()?
         .as_ptr()
-        .copy_to_nonoverlapping(buffer.cast::<u8>(), contentlength.min(size_to_read));
+        .copy_to_nonoverlapping(buffer.cast(), contentlength.min(size_to_read));
     let hd = HandleData::new(url, headers, None);
     // heap-allocate the struct and cast it to a void pointer so it can be passed around to PROJ
     let hd_boxed = Box::new(hd);
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Bump geo-types
+- Allow typeck to infer pointer types
 - Bump `approx` dev dependency to match `geo-types`. This doesn't affect
   downstream users, only those building the proj crate for development.
   -  <https://github.com/georust/proj/pull/138>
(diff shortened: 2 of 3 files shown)


src/network.rs
```diff
@@ -57,7 +57,7 @@ impl Drop for HandleData {
     // dereferencing it if need be so the resource is freed
     fn drop(&mut self) {
         if let Some(header) = self.hptr {
-            let _ = unsafe { CString::from_raw(header.as_ptr().cast::<i8>()) };
```
michaelkirk, line 60: So just to reiterate what's happening - c_char is defined differently depending on your platform. See the multiple definitions in [`std::os`](https://web.mit.edu/rust-lang_v1.25/arch/amd64_ubuntu1404/share/doc/rust/html/src/std/os/raw/mod.rs.html#34).

If we want it to work cross platform, we should be using the type alias (`c_char`) rather than the resolved type (i8 or u8). Or omit it altogether as you have here, using type inference, which also seems fine.

I can easily see how we got here...

[The source](https://doc.rust-lang.org/src/alloc/ffi/c_str.rs.html#397) uses the c_char type:
Screenshot 2023-06-26 at 11 30 40


Whereas [the rendered docs](https://doc.rust-lang.org/std/ffi/struct.CString.html#method.from_raw) use the resolved type alias:
Screenshot 2023-06-26 at 11 31 29

Presumably that's dependent on whatever arch the docs were targeting, so it will be inaccurate for people on those other platforms, and fail to compile. 

I'm not sure why that is, and I don't have time to investigate further now, but maybe it's: https://github.com/rust-lang/rust/issues/15823

michaelkirk (commented):
bors r+

src/network.rs
urschrei: Ahhh I saw that. `c_char` and was confused by it, because I treated the docs as canonical, like the fool that I am. Good catch.

PR #165, opened 2023-06-26
Prepare for 0.27.1 release

urschrei:
Is a point release OK here?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,9 +1,9 @@
 # Changes
 
-## Unreleased
+## 0.27.1
 
-- Bump geo-types
-- Allow typeck to infer pointer types
+- FIX: `network` feature fails to compile on aarch64/arm64
+  - https://github.com/georust/proj/issues/163
 - Bump `approx` dev dependency to match `geo-types`. This doesn't affect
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -1,6 +1,6 @@
 # Changes
 
-## Unreleased
+## 0.27.1
 
 - Bump geo-types
```
michaelkirk, line 5: nit: maybe delete this line and update the pre-existing "update to geo-types" a few lines down. It's kind of confusing to have both entries in the change log.

CHANGES.md
```diff
@@ -1,6 +1,6 @@
...
 - Allow typeck to infer pointer types
```
michaelkirk, line 6:
```suggestion
- FIX: `network` feature fails to compile on aarch64/arm64.
```

The actual change is kind of inscrutable - probably it's more useful to explain why you did it.

michaelkirk (approved):
>  Is a point release OK here?

Yes, I think that's what we want. 


Other than my changelog nits, LGTM!

(I choose to be optimistic, and think someone besides me, somewhere, enjoys reading change logs.)

PR #166, opened 2023-06-30
Update to libproj 9.2.1

urschrei:
(Draft for now because we'll have to update CI)


michaelkirk:
Want me to take a pass at updating CI?

urschrei:
I'm about to open a PR against CI, as it looks like your (delightful) regen script worked. It'd be great if you could have a quick look for sanity and then I can kick off the build / test / publish step, as I now have some more computing power so it might not be massively onerous…

urschrei:
https://github.com/georust/docker-images/pull/24

urschrei:
well, something is definitely Wrong with the new 1.70 image.

urschrei:
If I had to guess, it's because the 1.63 images are based on the image that was present on Docker Hub, and its architecture is `linux/amd64` whereas subsequent images including the 1.70 ones, are `linux/arm64/v8/` (I assume because I built them on an Apple M). Ugh.

michaelkirk:
Ah yes! I hadn’t really thought of it, but I’ve been doing builds on amd64.Want me to kick off a set of builds?On Jul 2, 2023, at 12:24, Stephan Hügel ***@***.***> wrote:﻿
If I had to guess, it's because the 1.63 images are based on the image that was present on Docker Hub, and its architecture is linux/amd64 whereas subsequent images including the 1.70 ones, are linux/arm64/v8/ (I assume because I built them on an Apple M). Ugh.

urschrei:
Yes but no rush! It's Sunday!

urschrei:
I'm confused about this failure.

michaelkirk (approved):
LGTM

urschrei:
bors r=michaelkirk

PR #169, opened 2023-07-14
Delete bors.toml in favor of GitHub merge queues

frewsxcv:


michaelkirk:
Screenshot 2023-07-27 at 08 08 58

🤔 It seems like it's merging directly to main and only running the tests after. Or am I confused?

urschrei:
Ughhhhhh I just replicated what we used in GeoJSON. Blast.

frewsxcv:
I wonder if it's because we have this specified:

```
  push:
    branches:
      - main
      - staging
      - trying
```

Which means CI won't run on other branches, like the temporary branch it creates for the queue

urschrei:
The other difference with geojson is that we aren't using an "aggregator" job the way we do in `proj` and `geo`…

urschrei:
That _shouldn't_ make a difference, but who knows.

michaelkirk:
I feel like this isn't supposed to be hard, but I can't figure out what we're doing wrong. 😭

michaelkirk:
Ohhhhh, well this was disabled. I've enabled it now.
Screenshot 2023-07-27 at 08 27 39

michaelkirk:
> well this was disabled. I've enabled it now.

That setting is actually confusing to me. It didn't solve the problem and I'm not sure what it does.

I think what we want is the "only merge non-failing pull requests" but that was already checked. 🤷 

Screenshot 2023-07-27 at 08 34 29

PR #176, opened 2023-08-06
Add ability to convert 3d coordinates. Fixes #150

### LINKED ISSUE #150: Need ability to convert 3D coordinates
`Proj::convert` assumes 2D coordinates, and passes `z: 0.0` into `proj_sys` (which handles 3D just fine). 

This is insufficient for conversions to/from (e.g.) ECEF coordinates, where the z coordinate is critical even for locations on the ground. 

Then need to access the private `c_proj` makes it impossible to define a trait with a minimally-modified `convert` method which takes a 3D point, passes all three values in/out of proj_sys, and returns a 3D point.

Some provision for 3D coords within this crate would be extremely helpful please :-)

PaulWagener:
This PR adds the ability to convert 3d coordinates.

The `Coord` trait is changed to add a Z coordinate so at least a minor version bump is warranted with this change.

### DIFF
--- a/src/proj.rs
+++ b/src/proj.rs
@@ -67,7 +67,8 @@ where
 {
     fn x(&self) -> T;
     fn y(&self) -> T;
-    fn from_xy(x: T, y: T) -> Self;
+    fn z(&self) -> T;
+    fn from_xyz(x: T, y: T, z: T) -> Self;
 }
 
 impl<T: CoordinateType> Coord<T> for (T, T) {
@@ -77,8 +78,26 @@ impl<T: CoordinateType> Coord<T> for (T, T) {
     fn y(&self) -> T {
         self.1
     }
-    fn from_xy(x: T, y: T) -> Self {
+    fn z(&self) -> T {
+        T::zero()
+    }
+    fn from_xyz(x: T, y: T, _z: T) -> Self {
         (x, y)
+    }
+}
+
+impl<T: CoordinateType> Coord<T> for (T, T, T) {
+    fn x(&self) -> T {
+        self.0
+    }
+    fn y(&self) -> T {
+        self.1
+    }
+    fn z(&self) -> T {
+        self.2
+    }
+    fn from_xyz(x: T, y: T, z: T) -> Self {
+        (x, y, z)
     }
 }
 
@@ -733,8 +752,10 @@ impl Proj {
         };
         let c_x: c_double = point.x().to_f64().ok_or(ProjError::FloatConversion)?;
         let c_y: c_double = point.y().to_f64().ok_or(ProjError::FloatConversion)?;
+        let c_z: c_double = point.z().to_f64().ok_or(ProjError::FloatConversion)?;
         let new_x;
         let new_y;
+        let new_z;
         let err;
         // Input coords are defined in terms of lambda & phi, using the PJ_LP struct.
         // This signals that we wish to project geodetic coordinates.
@@ -744,7 +765,7 @@ impl Proj {
         let coords = PJ_LPZT {
             lam: c_x,
             phi: c_y,
-            z: 0.0,
+            z: c_z,
             t: f64::INFINITY,
         };
         unsafe {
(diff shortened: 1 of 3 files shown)


urschrei:
Hi @PaulWagener, thanks for the PR!

As ever with 2D -> 3D, there's likely to be some discussion…

At minimum, I don't think we can remove the `xy` functions in one fell swoop as that would break a great deal of downstream code. I think marking them as deprecated in favour of the new ones is better, here.

As for `Coord`, we're looking at the same problem. In a sense, this is our fault for using an overly general name, but changing the signature will break a lot of code. How do people feel about a new `Coord3` (or maybe `Coord4` to include M values) struct and deprecating `Coord`.

Even more radically: we could adopt Rust-Geodesy's coordinate struct…

PaulWagener:
As long as 3D coordinates will be supported I will be happy :)

src/proj.rs
```diff
@@ -744,22 +765,24 @@ impl Proj {
...
             phi: c_y,
-            z: 0.0,
+            z: c_z,
             t: f64::INFINITY,
         };
         unsafe {
             proj_errno_reset(self.c_proj);
             // PJ_DIRECTION_* determines a forward or inverse projection
             let trans = proj_trans(self.c_proj, inv, PJ_COORD { lpzt: coords });
             // output of coordinates uses the PJ_XY struct
```
michaelkirk, line 775: Comment is out of date.

src/proj.rs
```diff
@@ -67,7 +67,8 @@ where
 {
     fn x(&self) -> T;
     fn y(&self) -> T;
-    fn from_xy(x: T, y: T) -> Self;
+    fn z(&self) -> T;
+    fn from_xyz(x: T, y: T, z: T) -> Self;
```
michaelkirk, line 71: Trait doc comments are now out of date.

michaelkirk:
@lx-88 - does using this branch work for you?

michaelkirk:
> How do people feel about a new Coord3 (or maybe Coord4 to include M values) struct and deprecating Coord.

A new **struct** or a new **trait**?

I thought the whole "bring your own struct" this was kind of nice so you don't have to first convert your data to an intermediate format. 


Just brainstorming... if avoiding api-breakage is a concern, could we introduce the new 3d method with a default impl?
```diff
 trait Coord {
   ...
+  #[deprecated(note = "implement from_xyz instead")]
   fn from_xy(x: f64, y: f64, z:f64) -> Self;

+  fn from_xyz(x: f64, y: f64, z:f64) -> Self {
+     Self::from_xy(x, y)
+  }
 }
```

nokonoko1203:
Hello!
I would be very happy if this crate could be made to correspond to three-dimensional coordinates!

I know that it is very difficult to do this, but what is the reason that this PR has not been merged?
Is there anything I can do to help?

balazsdukai:
In the meantime, while this PR is being worked on, here is a [fork](https://github.com/3DGI/tyler/blob/master/src/proj.rs) that reprojects 3d coordinates . [Here](https://github.com/3DGI/tyler/blob/85109336b206b566183cb36a9e57401a60132f99/src/formats.rs#L1267) is an example how I use it.
I did the minimum necessary for my use case, but it might be useful for others.

nokonoko1203:
This is great! I'll use it as a reference!
I was just thinking of using it for a similar project.
(PointCloud to 3D Tiles v1.1: https://github.com/MIERUNE/point-tiler)

arpadav:
I just encountered this lack-of-feature, and implemented the same solution, so I am glad this is getting attention. 

However, from an API standpoint, sometimes I just want to change my CRS for degrees of an ellipsoid with no 3d component, but sometimes I require 3d for operations like ECEF

So should users be required to add 0 z's for 2d coordinates in their `from_xyz` implementation, or should there be some sort of `CoordXy/Coord2` and `CoordXyz/Coord3` traits? 

I'm honestly fine with either solution, but am just curious on others thoughts

susu:
I've stumbled upon this missing functionality, trying to convert from ECEF (EPSG:4978) to EPSG:4326. Tried nokonoko1203's fork which works well (rebased to main). What leftover tasks remaining here to merge this?

nokonoko1203:
What is the work required to merge this PR?
I'm very happy if this feature is merged

PR #177, opened 2023-08-06
Fixes wrong coordinates in the doctest

PaulWagener:
They were off by a few meters but enough for the test to fail

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -171,8 +171,8 @@
 //!
 //! let result = proj.convert(donut_shop).unwrap();
 //!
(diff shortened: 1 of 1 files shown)


lnicola:
I don't know if this is actually necessary:

```
$ echo "34.095620 -118.283555" | cs2cs EPSG:4326 EPSG:3309
158458.67	-434296.88 0.00
$ echo "34.095620 -118.283555" | PROJ_NETWORK=ON cs2cs EPSG:4326 EPSG:3309
158460.19	-434297.05 0.00
```

Maybe you have the grids available on your system.

PaulWagener:
You're right. Both commands result in `158460.19 -434297.05 0.00` for me

PR #188, opened 2024-01-17
Bump MSRV to 1.70

urschrei:


lnicola (approved):

lnicola:
Let's also change it in https://github.com/georust/proj/blob/main/Cargo.toml?

PR #190, opened 2024-01-26
Depend on `libsqlite3-sys` for bundled builds

weiznich:
This commit adds an optional dependency to libsqlite3-sys to provide a bundled version of libsqlite3 as well instead of relying on a system provided version.


Cargo.toml
```diff
@@ -38,3 +38,7 @@ geo-types = { version = "0.7.10", features = ["approx"] }
 [package.metadata.docs.rs]
 features = [ "proj-sys/nobuild", "network", "geo-types" ]
 rustdoc-args = ["--cfg", "docsrs"]
+
+
+[patch.crates-io]
+libsqlite3-sys = { git = "https://github.com/rusqlite/rusqlite/", rev = "42a82fb" }
```
weiznich, line 44: This currently relays on an accepted but unreleased patch to the libsqlite3-sys crate: https://github.com/rusqlite/rusqlite/pull/1441

weiznich: This is now published as libsqlite-sys 0.28, therefore I just updated the version and removed that patch

weiznich:
What's required to move this forward?

urschrei:
Could you address the CI failures?

weiznich:
That's an interesting failure mode. It seems like the build script can decide at runtime to fall back building the libproj. That's problematic because we would need to know beforehand to whether to depend on libsqlite-sys or not. I see three solutions:

* Remove the fallback and require setting a feature flag for bundled builds
* Always depend and libsqlite-sys and always build the bundled variant of that library. I think we can say that we don't want to link libsqlite to the final binary, but the build overhead will always be there
* Only use that approach when gathering explicit feature flag is set, for the fallback builds conserve the old behaviour (relying on a system provided libsqlite3)

I'm not sure which way would be preferred.

weiznich:
I've pushed an update that should pass CI.

urschrei:
Great, thank you! I had no idea `link-cplusplus` existed. I don't see any obvious downsides to this (and it may also get us a bit closer to wasm builds?), so I'm happy to merge once at least one of the other two requested reviewers has a look.

proj-sys/Cargo.toml
```diff
@@ -12,6 +12,8 @@ links = "proj"
 rust-version = "1.70"
 
 [dependencies]
+libsqlite3-sys = { version = "0.28", features = ["bundled"] }
```
michaelkirk, line 15: Based on your description, I was expecting this to be an optional dependency. What am I misunderstanding?

weiznich: It's unfortunately not possible to keep that as optional dependency as the build script can decide at runtime to build libproj from source. See [here](https://github.com/georust/proj/pull/190#issuecomment-1936465465) for more details.

weiznich: @michaelkirk Does this answer your question? Or would you like something to be changed here?

michaelkirk: Well ideally we wouldn't be compiling a thing that we don't need. I understand that it's complicated though.

Will `libsqlite3-sys` build sqlite3 even if it's already been installed on the system (e.g. via a package manager)?

weiznich: @michaelkirk Yes with the `bundled` feature flag enabled `libsqlite3-sys` will always build `libsqlite3` from source and statically link it. They do not provide a option that performs runtime detection whether the crate is installed or not and that falls back to compiling if the crate is not found. That means with the current version we will always build `libsqlite3` from source, but we will only link that version if we also build libproj from source. 

That written: I had another idea how we could maybe solve that problem in a better way. We could not enable the `bundled` feature flag in `proj-sys` at all. We can then use the `DEP_SQLITE3_*` variables to check at runtime whether or not the `libsqlite3-sys` crate build a libsqlite3 version from source by checking if these environment variables exist. That would allow users to control how libsqlite3 should be linked, as they then can control the `bundled` flag there on their own as feature flags are additive. If these variables are not set we would fall back to trying to get a system version of libsqlite3 for the build, which is the current behavior. If that sounds more acceptable I would change the PR to use that approach instead.

weiznich: @michaelkirk I've pushed https://github.com/georust/proj/pull/190/commits/44a503869709ea793a5f8452bfb9e2a99b7160fc which essentially implements what was described in my previous comment. Is there anything else required? I would like to move this forward to unblock the [gdal bundling PR](https://github.com/georust/gdal/pull/517)

weiznich:
Is there any interest in moving this PR forward from the relevant maintainers?

proj-sys/Cargo.toml
urschrei: @weiznich Just so I'm clear: there's now no unneeded sqlite build unless the `bundled` feature is enabled?

weiznich: Yes, I changed it that way that it only uses the statically build libsqlite3 version if the user specified somewhere else in their dependency tree that libsqlite3 should be bundled.

urschrei: > uses the statically build

So it still builds sqlite every time, but discards the static build if it's not needed?

weiznich: No, it doesn't configure the bundled build for sqlite at all.

If we reach the stage where we build `libproj` from source is just checks if someone else has configured that `libsqlite3-sys` should build libsqlite3 from source and if that's the case it will use these build artifacts. That check is done by these two `DEP_*` variables, which are set by `libsqlite3-sys` for the bundled build. For that to work `proj-sys` needs to depend on `libsqlite3-sys` (otherwise we won't see the variables), but that shouldn't be a problem as that only runs the build script of that crate, which by default will only handle that libsqlite3 is linked to the binary, not that it is build from source.

urschrei: Great, thanks for clarifying!

lnicola (approved):

lnicola:
The interaction between this and `libsqlite3-sys` seems quite subtle, but probably the best solution we can find. It might be worth adding a comment about it?

> but I'm hesitant to merge unless it's at least likely that it'll land

I hope the GDAL PR lands. It doesn't have the drivers I use most, but it can be expanded by making `-src` crates for those in the future. And it's really useful today, especially for vector processing and users who are fine with the basic GeoTIFF support.

michaelkirk:
I think it's weird to build sqlite3 even though we might not use it. It's not a huge problem though, and apparently this is unavoidable. I'm neutral on this PR. If someone wants to advocate for merging it that's fine with me.

urschrei:
> I think it's weird to build sqlite3 even though we might not use it. It's not a huge problem though, and apparently this is unavoidable. I'm neutral on this PR. If someone wants to advocate for merging it that's fine with me.

I think that we now _don't_ build it if it's not required:

https://github.com/georust/proj/pull/190#discussion_r1547636023

weiznich:
> >    I think it's weird to build sqlite3 even though we might not use it. It's not a huge problem though, and apparently this is unavoidable. I'm neutral on this PR. If someone wants to advocate for merging it that's fine with me.
>
> I think that we now don't build it if it's not required:

It's a bit confusing but it's essentially as follows:

* We always build the rust crate `libsqlite3-sys`
* Downstream users might configure `libsqlite3-sys` to build the c library libsqlite3, in which case we link that library

proj-sys/build.rs
```diff
@@ -99,6 +99,20 @@ fn build_from_source() -> Result<std::path::PathBuf, Box<dyn std::error::Error>>
     config.define("BUILD_PROJSYNC", "OFF");
     config.define("ENABLE_CURL", "OFF");
 
+    // we check here whether or not these variables are set by cargo
+    // if they are set `libsqlite3-sys` was build with the bundled feature
```
michaelkirk, line 103:
```suggestion
    // if they are set, `libsqlite3-sys` was built with the bundled feature
```

proj-sys/build.rs
```diff
@@ -99,6 +99,20 @@ fn build_from_source() -> Result<std::path::PathBuf, Box<dyn std::error::Error>>
...
+    // enabled, which in turn allows us to relay on the build libsqlite3 version
```
michaelkirk, line 104:
```suggestion
    // enabled, which in turn allows us to rely on the built libsqlite3 version
```

proj-sys/build.rs
```diff
@@ -99,6 +99,20 @@ fn build_from_source() -> Result<std::path::PathBuf, Box<dyn std::error::Error>>
...
+    // and link it statically
+    //
+    // If these are not set it's necessary that libsqlite3 exists on the build system
```
michaelkirk, line 107:
```suggestion
    // If these are not set, it's necessary that libsqlite3 exists on the build system
```

proj-sys/build.rs
```diff
@@ -99,6 +99,20 @@ fn build_from_source() -> Result<std::path::PathBuf, Box<dyn std::error::Error>>
...
+    // in a location accecible by cmake
```
michaelkirk, line 108:
```suggestion
    // in a location accessible by cmake
```

proj-sys/src/lib.rs
```diff
@@ -28,6 +28,11 @@
 //! implement your own set of callbacks if you wish to make use of them (see the
 //! [`proj`](https://crates.io/crates/proj) crate for an example).
 
+#[cfg(bundled_build)]
+extern crate libsqlite3_sys;
```
michaelkirk, line 32: Why do we need these `extern crate` declarations?

michaelkirk:
> It's a bit confusing but it's essentially as follows:
>
> We always build the rust crate libsqlite3-sys
> Downstream users might configure libsqlite3-sys to build the c library libsqlite3, in which case we link that library

Ah, actually that resolves my concerns. Thanks for your patience @weiznich!

Some small typos and a question about the use of `extern`, but otherwise LGTM

proj-sys/src/lib.rs
weiznich: Rustc only links crates that are explictly used somewhere, either by actually use items from that crate or by having an explicit `extern crate …` statement somewhere. So this essentially ensures that we actually link `libsqlite3` (the c library) for the bundled build case. We set the cfg flag from the build script if we actually build libproj from source and statically link it. Only then we need to also link libsqlite3.

weiznich:
@michaelkirk Is there anything that I can do to move this PR forward?

proj-sys/build.rs
```diff
@@ -72,6 +72,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 // returns the path of "include" for the built proj
 fn build_from_source() -> Result<std::path::PathBuf, Box<dyn std::error::Error>> {
     eprintln!("building libproj from source");
+    println!("cargo:rustc-cfg=bundled_build");
```
michaelkirk, line 75: What is this used for?

michaelkirk (approved):
Sorry for letting this slip and thanks for the prod @weiznich.

I'm OK to merge if @urschrei is.

proj-sys/build.rs
weiznich: It conditionally enables this `extern crate` statement here: https://github.com/georust/proj/pull/190/files#diff-bdc7b35321c79a3531cffc03776c83af67e9481bf67db7f7a94540dd559b0036R32, which in turn conditionally controls whether or not we want to link (statically or dynamically) libsqlite3 to the final binary.

michaelkirk: My memory is short and my ctrl+f is weak. Thank you.

urschrei:
I'm happy to merge!

michaelkirk:
The CI failure seems spurious - the floating point on macos are off by a very small amount. It might be related to GH switching to arm64 (macos-14-arm64). I'll open a PR momentarily to address the tests and then we can re-run this one.

PR #192, opened 2024-01-30
Add bundling support on windows

weiznich:
We need to always build release builds there (to prevent linker errors due to mixing up debug and release runtimes) and we need an additional flag to prevent the static library to expect some proj.dll file to exist

### DIFF
--- a/proj-sys/build.rs
+++ b/proj-sys/build.rs
@@ -121,6 +121,15 @@ fn build_from_source() -> Result<std::path::PathBuf, Box<dyn std::error::Error>>
         config.define("ENABLE_TIFF", "OFF");
     }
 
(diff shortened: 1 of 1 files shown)


weiznich:
@urschrei @michaelkirk @lnicola What is required to move this forward? This is required for windows support of https://github.com/georust/gdal/pull/517

urschrei (approved):
I can't test this on msvc myself, but this seems uncontroversial to me.

urschrei:
Sorry for the delay on this @weiznich. I need to figure out why the CI check hasn't run – could you rebase against `main`?

lnicola (approved):
Sorry, I was away last week. I'm not a fan of how this looks (wrapping, capitalization), but the change makes sense.

PR #193, opened 2024-02-21
Allow creating a Proj from an owned String

TomFryersMidsummer:
`CString::new` takes any `Into<Vec<u8>>`, which includes `String`. This should make it possible to avoid reallocating when passing in an owned `String` with sufficient capacity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ use proj::Proj;
 
 let from = "EPSG:2230";
 let to = "EPSG:26946";
-let ft_to_m = Proj::new_known_crs(&from, &to, None).unwrap();
+let ft_to_m = Proj::new_known_crs(from, to, None).unwrap();
 let result = ft_to_m
     .convert((4760096.421921f64, 3744293.729449f64))
     .unwrap();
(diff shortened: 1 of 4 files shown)


TomFryersMidsummer:
I think the `TryFrom` implementation extension doesn't cause breakage, since that never accepted `&&str`.

michaelkirk:
I think the existing type signature is going to be more intuitive for the caller. If I'm understanding your motivation, having to type in a `&` before your owned string does not seem like a big enough hurdle to justify this change.

Unless I'm misunderstanding, I don't think we should merge this.

michaelkirk:
Sorry, I missed your initial comment that this is about avoiding re-allocation inside of CString.  But my conclusion is the same - it seems unlikely that gains here will be worth having a more opaque/confusing input type. 

Are you experiencing this part of the code as a bottleneck?

TomFryersMidsummer:
I just find it a bit odd for `Proj` to ask for a reference if in the end it wants to own its input. (See [C-CALLER-CONTROL](https://rust-lang.github.io/api-guidelines/flexibility.html#caller-decides-where-to-copy-and-place-data-c-caller-control).)

Having had another look, I don't think this is performance-critical for us, so I don't mind if you don't think it's worth it.

The type could also be `Into<String>`, which is clearer and more type-safe, but less flexible.

michaelkirk:
I appreciate the attempt, but I think on balance this is not worthwhile, so closing.

PR #196, opened 2024-04-01
Update proj-sys to use libproj 9.4.0

urschrei:


michaelkirk:
re: macos failure

> ==> Pouring proj--9.3.1.monterey.bottle.tar.gz

But the latest version available on hombrew is 9.4.0 https://formulae.brew.sh/formula/proj

Do we need to `brew update` to refresh the index or something?

.github/workflows/test.yml
```diff
@@ -141,6 +141,7 @@ jobs:
     runs-on: macos-latest
     env:
       _PROJ_SYS_TEST_EXPECT_BUILD_FROM_SRC: 0
+      HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK: TRUE
```
michaelkirk, line 144: Wow, this variable name is a doozy. (I realize it's from the homebrew project - not something you came up with 🤷‍♂️).

Is the idea that since we're manually updating we can ignore the auto-update feature?

michaelkirk (approved):
LGTM

.github/workflows/test.yml
urschrei: Yes, as it transiently causes the build to fail as those dependent package installs are randomly failing for no reason (see last two CI failures on macos)

PR #198, opened 2024-04-04
Switch to GHA / GHCR Docker images

urschrei:


.github/workflows/test.yml
```diff
@@ -15,6 +15,19 @@ env:
   RUST_LATEST: "1.77"
 
 jobs:
+  # work around GHA env strictness
```
michaelkirk, line 18: I'm not worried about it, but I'm curious, what's it about?

michaelkirk (approved):
LGTM

.github/workflows/test.yml
urschrei: You can't use variables defined in the env section to build matrix entries (and some other places), I assume for security reasons, though I can't see how anyone can inject something here. Found the problem by linting with https://rhysd.github.io/actionlint/ which actually explains the problem (GHA just fails with a don't know what env is error)

PR #199, opened 2024-04-23
update tests for arm64

michaelkirk:


src/lib.rs
```diff
@@ -171,8 +171,8 @@
 //!
 //! let result = proj.convert(donut_shop).unwrap();
 //!
-//! assert_relative_eq!(result.x(), 158458.67, epsilon=1e-2);
-//! assert_relative_eq!(result.y(), -434296.88, epsilon=1e-2);
+//! assert_relative_eq!(result.x(), 158460.19, epsilon=1e-2);
+//! assert_relative_eq!(result.y(), -434297.05, epsilon=1e-2);
```
michaelkirk, line 175: The other changes don't really concern me, but this one feels like a somewhat large change. I'm not sure why.

michaelkirk: Ubuntu x86_64 value: 158458.67251293262
macos14 arm64 value:  158460.18744483538

urschrei: I have no idea why the difference would be this pronounced. Hmm!

michaelkirk: Apparently it's a local only issue. The GH runners are close (enough) to the recorded values. I'm not sure what's gong on, but not really worried about it if it's only my machine.

(lazy web: how do I clear my downloaded grid cache?)

frewsxcv (approved):

michaelkirk:
failure: 

> warning: spurious network error (3 tries remaining): [7] Couldn't connect to server (Failed to connect to index.crates.io port 443 after 4002 ms: Couldn't connect to server)

retrying...

PR #204, opened 2024-05-31
Add Proj::id_code and Proj::id_auth_name 

zwishing:
Add method `to_epsg` for Proj, return epsg code

### DIFF
--- a/proj-sys/Cargo.toml
+++ b/proj-sys/Cargo.toml
@@ -12,7 +12,7 @@ links = "proj"
 rust-version = "1.70"
 
 [dependencies]
-libsqlite3-sys = "0.28"
+libsqlite3-sys = ">=0.28,<0.31"
 link-cplusplus = "1.0"
 
 [build-dependencies]


src/proj.rs
```diff
@@ -584,6 +584,17 @@ impl Proj {
         transform_epsg(ctx, from, to, area)
     }
 
+    /// Returns epsg code for Proj
+    /// 
+    /// # Safety
+    /// This method contains unsafe code.
+    pub fn to_epsg(&self) -> Result<&str, str::Utf8Error> {
```
lnicola, line 591: I'd rather call it `identifier` or `code` or `id_code` or `authority_code`. Not all of them are EPSG.

Zwishing: Thank you for your suggestion. I have revised the PR accordingly。

src/proj.rs
```diff
@@ -1,20 +1,8 @@
...
-    proj_grid_cache_set_enable, proj_info, proj_normalize_for_visualization, proj_pj_info,
-    proj_trans, proj_trans_array, proj_trans_bounds, PJconsts, PJ_AREA, PJ_CONTEXT, PJ_COORD,
-    PJ_DIRECTION_PJ_FWD, PJ_DIRECTION_PJ_INV, PJ_INFO, PJ_LPZT, PJ_XYZT,
-};
-use std::{
-    convert, ffi,
-    fmt::{self, Debug},
-    str,
-};
+use proj_sys::{proj_area_create, proj_area_destroy, proj_area_set_bbox, proj_cleanup, proj_context_create, proj_context_destroy, proj_context_errno, proj_context_get_url_endpoint, proj_context_is_network_enabled, proj_context_set_search_paths, proj_context_set_url_endpoint, proj_create, proj_create_crs_to_crs, proj_destroy, proj_errno_string, proj_get_area_of_use, proj_get_id_code,proj_get_id_auth_name, proj_grid_cache_set_enable, proj_info, proj_normalize_for_visualization, proj_pj_info, proj_trans, proj_trans_array, proj_trans_bounds, PJconsts, PJ, PJ_AREA, PJ_CONTEXT, PJ_COORD, PJ_DIRECTION_PJ_FWD, PJ_DIRECTION_PJ_INV, PJ_INFO, PJ_LPZT, PJ_XYZT};
```
michaelkirk, line 4: In general when you see a big diff that's a hint that formatting might be off. Please run `cargo fmt` on your branch.

Zwishing: OK, new PR has `cargo fmt`

src/proj.rs
```diff
@@ -592,6 +593,28 @@ impl Proj {
...
     }
 
+    /// Return Authority code for Proj
+    ///
+    /// # Safety
+    /// This method contains unsafe code.
+    pub fn id_code(&self) -> Result<&str, str::Utf8Error> {
+        let pj = self.c_proj as *const PJ;
+        let c_char = unsafe { proj_get_id_code(pj, 0) };
+        let c_str = unsafe { CStr::from_ptr(c_char) };
```
rouault, line 603: I believe there should be nullptr checking here as proj_get_id_code() might return null in case of error (typically if the CRS is custom and not tied to a known code), and https://doc.rust-lang.org/std/ffi/struct.CStr.html#method.from_ptr indicates the CStr::from_ptr() shouldn't be called with a null ptr

src/proj.rs
```diff
@@ -592,6 +593,28 @@ impl Proj {
...
+    }
+
+    /// Return Authority Name for Proj
+    ///
+    /// # Safety
+    /// This method contains unsafe code.
+    pub fn id_auth_name(&self) -> Result<&str, str::Utf8Error> {
+        let pj = self.c_proj as *const PJ;
+        let c_char = unsafe { proj_get_id_auth_name(pj, 0) };
+        let c_str = unsafe { CStr::from_ptr(c_char) };
```
rouault, line 614: same as above

src/proj.rs
Zwishing: When testing, I found that the Proj created by proj4str and esriwkt could not recognize the code and name。

rouault: > I found that the Proj created by proj4str and esriwkt could not recognize the code and name。

Yes, that's fully expected. Not all CRS representations contain an authority name and code

Zwishing: If there are corresponding ideas or methods, converting them would be ideal.

rouault: you can try using proj_identify() to find a known CRS that is identical to the current one, but there's no guarantee it will succeed. Custom CRS with custom projection parameter values are fully valid

Issue #189, opened 2024-01-24
Investigate the possibility of replacing reqwest with ureq


urschrei:
Even though it's behind a feature gate, reqwest has an extensive dependency tree which has an impact on compile times. We should have a look at https://github.com/algesten/ureq as an alternative.  We don't do anything especially exotic with reqwest; the main issue will be figuring out how headers are handled.

nk9:
Given that this has been merged, can this be closed?

PR #205, opened 2024-06-12
Port from reqwest to ureq

TomFryersMidsummer:
Addresses #189, reducing `cargo tree -F network | wc -l` from 296 to 190.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ geo-types = { version = "0.7.10", optional = true }
 libc = "0.2.119"
 num-traits = "0.2.14"
 thiserror = "1.0.30"
-reqwest = { version = "0.12.0", optional = true, default-features = false, features = ["blocking", "rustls-tls"] }
+ureq = { version = "2.0.0", optional = true }
 
 [workspace]
 members = ["proj-sys"]
@@ -28,7 +28,7 @@ members = ["proj-sys"]
 default = ["geo-types"]
 bundled_proj = [ "proj-sys/bundled_proj" ]
 pkg_config = [ "proj-sys/pkg_config" ]
-network = ["reqwest", "proj-sys/network"]
+network = ["ureq", "proj-sys/network"]
 
 [dev-dependencies]
 # approx version must match the one used in geo-types
--- a/src/network.rs
+++ b/src/network.rs
@@ -180,10 +180,9 @@ unsafe fn _network_open(
     // hand the response off to the error-handler, continue on success
     error_handler(&mut res, in_case_of_error)?;
     // Write the initial read length value into the pointer
-    let Some(Ok(contentlength)) = res.header("Content-Length").map(str::parse) else {
+    let Some(Ok(contentlength)) = res.header("Content-Length").map(str::parse::<usize>) else {
         return Err(ProjError::ContentLength);
     };
-    out_size_read.write(contentlength);
     let headers = res
         .headers_names()
         .into_iter()
@@ -200,6 +199,7 @@ unsafe fn _network_open(
     res.into_reader()
         .take(size_to_read as u64)
         .read_to_end(&mut buf)?;
+    out_size_read.write(buf.len());
     buf.as_ptr().copy_to_nonoverlapping(buffer.cast(), capacity);
     let hd = HandleData::new(url, headers, None);
     // heap-allocate the struct and cast it to a void pointer so it can be passed around to PROJ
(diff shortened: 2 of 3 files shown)


urschrei:
Thanks for this @TomFryersMidsummer! As I note in the issue, I don't consider the proj network use case as requiring async http, so reqwest was not a conscious choice and I'm happy to see it replaced.

urschrei (approved):

src/network.rs
```diff
@@ -327,26 +335,36 @@ fn _network_read_range(
...
+    let Some(Ok(contentlength)) = res.header("Content-Length").map(str::parse::<usize>) else {
+        return Err(ProjError::ContentLength);
+    };
     // Copy the downloaded bytes into the buffer so it can be passed around
+    let capacity = contentlength.min(size_to_read);
+    let mut buf = Vec::with_capacity(capacity);
+    res.into_reader()
+        .take(size_to_read as u64)
+        .read_to_end(&mut buf)
+        .map_err(|_| ProjError::ReadError)?;
```
lnicola, line 364: I think we should propagate the error here, like for `NetworkError`.

src/proj.rs
```diff
@@ -107,24 +107,31 @@ pub enum ProjError {
     Network,
     #[error("Could not set remote grid download callbacks")]
     RemoteCallbacks,
-    #[error("Couldn't build request")]
+    #[error("Couldn't access the network")]
     #[cfg(feature = "network")]
-    BuilderError(#[from] reqwest::Error),
+    NetworkError(Box<ureq::Error>),
```
lnicola, line 112: Let's mark `ProjError` as `#[non_exhaustive]`, so we don't take a breaking change next time we add a variant.

src/network.rs
```diff
@@ -327,26 +335,36 @@ fn _network_read_range(
...
+                let v = res.header(&h)?.to_string();
+                (h, v)
+            })
+        })
+        .collect();
+    let Some(Ok(contentlength)) = res.header("Content-Length").map(str::parse::<usize>) else {
+        return Err(ProjError::ContentLength);
+    };
     // Copy the downloaded bytes into the buffer so it can be passed around
+    let capacity = contentlength.min(size_to_read);
```
lnicola, line 359: This is a pre-existing issue, but we aren't consistent about `out_size_read`. If `Content-Length` is larger than `size_to_read`, we set `out_size_read` to a value larger than the actual length of the buffer. We should just set it to `buf.len()`, regardless or how we handle that case.

Same for `_network_open` of course.

lnicola (approved):
LGTM with some nits.

I'm not convinced that `ureq` is the better choice, long-term (vs. `reqwest` or `isaahc` / `libcurl`), but I guess it's pretty easy to revert in the future.

Cargo.toml
```diff
@@ -19,7 +19,7 @@ geo-types = { version = "0.7.10", optional = true }
 libc = "0.2.119"
 num-traits = "0.2.14"
 thiserror = "1.0.30"
-reqwest = { version = "0.12.0", optional = true, default-features = false, features = ["blocking", "rustls-tls"] }
+ureq = { version = "2.0.0", optional = true }
```
michaelkirk, line 22: Did you actually test this on 2.0.0? Otherwise please specify whatever version you tested on (latest is 2.9.7).

Otherwise it's hard to know if you are leveraging any functionality that was added in a minor release.

michaelkirk (commented):
Thanks for the PR. I agree with @lnicola's feedback as well.

Cargo.toml
TomFryersMidsummer: Yes, I tested it on 2.0.0.

TomFryersMidsummer:
Thank you for the feedback. I believe I've addressed everything now

lnicola:
I believe `_network_read_range` still returns the Content-Lengt instead of the actual size of the data.

lnicola:
Looks good, at least looking at the new commits, but I'll let someone else merge it since I've never used this crate.

michaelkirk (approved):

frewsxcv (approved):
I prefer `reqwest` over `ureq` but I also don't use this feature of `proj` and hope we can just remove it someday soooo here's a ✅

urschrei:
> I also don't use this feature of `proj` and hope we can just remove it someday

The network feature? Hard to imagine it going anywhere: the grid download functionality is a big step forward; without it you have to ship a couple of hundred MBs around with every libproj (which we wouldn't be able to do bc of crates.io limitations, even if we wanted to), and older versions have out-of-date grids leading to less-accurate conversions etc.

frewsxcv:
Oops maybe I'm mixing up the network functionality in a different crate, disregard!

PR #209, opened 2024-09-27
Allow to use different versions of the libsqlite3-sys crate

weiznich:
I've opted to allow several different versions there instead of just bumping the supported version as that makes it much easier to use proj-sys in a complex project where other crates depend on libpsqlite3-sys as well. Cargo then will figure out the right version for all dependencies as long as there is a overlapping version.


urschrei (approved):
LGTM

PR #218, opened 2025-01-20
Update libsqlite3-sys to allow 0.31 as well

weiznich:
---

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -69,27 +69,37 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        container_image:
+        rust_version:
           # Minimum supported rust (MSRV)
-          - ghcr.io/georust/proj-ci:proj-${{ needs.compute.outputs.LIBPROJ_VERSION }}-rust-${{ needs.compute.outputs.RUST_MSRV }}
+          - ${{ needs.compute.outputs.RUST_MSRV }}
           # Latest stable rust
-          - ghcr.io/georust/proj-ci:proj-${{ needs.compute.outputs.LIBPROJ_VERSION }}-rust-${{ needs.compute.outputs.RUST_LATEST }}
+          - ${{ needs.compute.outputs.RUST_LATEST }}
         features:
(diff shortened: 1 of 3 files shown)


weiznich:
Seems like CI fails due to the new libsqlite3 version requiring a newer rust version

lnicola:
We've just bumped the MSRV, I'm all for it.

urschrei:
What would our new MSRV need to be? `geo` is "officially" current - 2 right?

weiznich:
Technically this wouldn't need a msrv change, as proj still would support the older `libsqlite3-sys` versions, so anyone that depends on older rust versions could just use an older libsqlite3-sys version. The question is more how do you want to express that in your CI setup.

weiznich:
It seems like it's currently impossible for me to work on the CI as it's not possible to run jobs here without approval and I also cannot run jobs on a fork as its [impossible to pull images there](https://github.com/GiGainfosystems/proj/pull/1)

urschrei:
Your PR runs here should no longer require approval. We're aware of the image-pull annoyance, I just haven't had time to figure out how to make the images public so they work in forks.

urschrei:
@weiznich your fork should now be able to run its Actions. lmk if it can't.

weiznich:
It should be ready to review now.

lnicola (approved):
TBH I would have gone for a MSRV bump instead of the extra complexity, but I guess it looks fine otherwise

weiznich:
Given that the 2024 edition will make `resolver = "3"` the default case + that resolver will use declared supported rust versions of dependencies it's meaningful in my opinion to test for the minimal supported versions of dependencies to make sure that this not accidentally breaks under `resolver="3"`

urschrei:
FYI (reviewers, future selves etc) this is the crate being used in the new action in the PR: https://github.com/taiki-e/cargo-hack

urschrei:
@lnicola wdyt? I'm inclined to merge this.

lnicola:
Yeah, should be good.

PR #219, opened 2025-01-24
Add explicit auth to CI

urschrei:
I have no idea whether this will work…


urschrei:
It [does not](https://github.com/urschrei/rust-proj/actions/runs/12551140835).

PR #221, opened 2025-03-07
Bump libsqlite3-sys to support version 0.32 as well

weiznich:
This commit increases the maximal supported libsqlite3-sys version to include 0.32 as well. It also fixes a warning about a unknown cfg by just adding it to the allow list as we set that via the build script.

---

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 - Update build status badge
 - Update MSRV to 1.81
+- Add support for building with libsqlite3-sys 0.32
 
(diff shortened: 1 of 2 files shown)


weiznich:
The CI errors do not seem to be related to the libsqlite3-sys version bump itself, but to some other dependency updates.

urschrei:
This is due to a dependency of ureq (litemap) having an msrv of 1.81

urschrei:
@weiznich We've resolved the build error. Please rebase against main so we can merge this.

urschrei (approved):

PR #222, opened 2025-03-15
Update the MSRV to 1.81 and the edition to 2021

urschrei:
- [ ] I added an entry to the project's change log file if knowledge of this change could be valuable to users.
---


urschrei:
This is required in order to unblock #221. We can then release, and unblock geo by bumping its proj dependency.

lnicola (approved):

PR #223, opened 2025-03-15
Update to libproj 9.6.0

urschrei:
---

This will need new Docker images, and for #222 to merge before it can be merged.


DEVELOPMENT.md
```diff
@@ -4,7 +4,7 @@
 Bindings were generated with the following command:
 
 ```sh
-bindgen --distrust-clang-mangling --blocklist-type max_align_t wrapper.h -- -I  PROJSRC/proj-9.4.0/src
+bindgen --distrust-clang-mangling --blocklist-type max_align_t wrapper.h -- -I  PROJSRC/proj-9.6.0/src
```
lnicola, line 7: Do we still need `--distrust-clang-mangling`? It seems to be a workaround for some Clang version from 9 years ago.

lnicola (approved):

DEVELOPMENT.md
urschrei: Good question – I added it way back in 476132ee, but let me see if we still need it.

PR #226, opened 2025-04-14
Add coordinate metadata creation and query functions

urschrei:
---

### DIFF
--- a/src/proj.rs
+++ b/src/proj.rs
@@ -3,12 +3,14 @@ use libc::{c_char, c_double};
 use num_traits::Float;
 use proj_sys::{
     proj_area_create, proj_area_destroy, proj_area_set_bbox, proj_as_projjson, proj_cleanup,
-    proj_context_create, proj_context_destroy, proj_context_errno, proj_context_get_url_endpoint,
-    proj_context_is_network_enabled, proj_context_set_search_paths, proj_context_set_url_endpoint,
-    proj_create, proj_create_crs_to_crs, proj_destroy, proj_errno_string, proj_get_area_of_use,
-    proj_grid_cache_set_enable, proj_info, proj_normalize_for_visualization, proj_pj_info,
-    proj_trans, proj_trans_array, proj_trans_bounds, PJconsts, PJ_AREA, PJ_CONTEXT, PJ_COORD,
-    PJ_DIRECTION_PJ_FWD, PJ_DIRECTION_PJ_INV, PJ_INFO, PJ_LPZT, PJ_XYZT,
+    proj_context_clone, proj_context_create, proj_context_destroy, proj_context_errno,
+    proj_context_get_url_endpoint, proj_context_is_network_enabled, proj_context_set_search_paths,
+    proj_context_set_url_endpoint, proj_coordinate_metadata_create,
+    proj_coordinate_metadata_get_epoch, proj_create, proj_create_crs_to_crs, proj_destroy,
+    proj_errno_string, proj_get_area_of_use, proj_grid_cache_set_enable, proj_info,
+    proj_normalize_for_visualization, proj_pj_info, proj_trans, proj_trans_array,
+    proj_trans_bounds, PJconsts, PJ_AREA, PJ_CONTEXT, PJ_COORD, PJ_DIRECTION_PJ_FWD,
+    PJ_DIRECTION_PJ_INV, PJ_INFO, PJ_LPZT, PJ_XYZT,
 };
 use std::ptr;
 use std::{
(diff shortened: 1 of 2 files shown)


src/proj.rs
```diff
@@ -1290,6 +1338,14 @@ mod test {
     }
 
     #[test]
+    fn test_metadata_creation() {
+        let wgs84 = "EPSG:4326";
+        let epoch = 2021.3;
+        let proj = Proj::new(wgs84).unwrap();
+        let np = proj.coordinate_metadata_create(epoch).unwrap();
+        assert_eq!(np.coordinate_metadata_get_epoch(), 2021.3);
```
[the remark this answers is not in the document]
urschrei: `libproj` uses F64 values to store epochs, denoted by e.g. `year.month`. No arithmetic is performed using epochs, so I think this is fine.

urschrei:
I think this will need a corresponding ProjBuilder method to work with e.g. the network features, which means (I think) adding a c_proj field.

src/proj.rs
```diff
@@ -498,6 +500,56 @@ pub struct Proj {
...
+    /// This retrieves the epoch associated with a coordinate metadata object.
+    ///
+    /// # Returns
+    ///
+    /// The epoch value as a float, or [`f64::NAN`] if the object doesn't have an associated epoch.
+    ///
+    /// # Safety
+    /// This method contains unsafe code.
+    pub fn coordinate_metadata_get_epoch(&self) -> f64 {
+        unsafe { proj_coordinate_metadata_get_epoch(self.ctx, self.c_proj) }
```
michaelkirk, line 550: Have you had any more luck than me (zero) finding documentation for these methods?

https://github.com/search?q=repo%3AOSGeo%2FPROJ%20proj_coordinate_metadata_get_epoch&type=code

Is `epoch` in any particular units or are those units specific to each projection?

src/proj.rs
```diff
@@ -498,6 +500,56 @@ pub struct Proj {
...
 
 impl Proj {
+    /// Create a coordinate metadata object to be used in coordinate operations.
+    ///
+    /// This creates a coordinate metadata object that can be used in coordinate operations,
+    /// such as transformations. The coordinate metadata object contains information
+    /// about the coordinates, such as the epoch they are referenced to.
+    ///
+    /// # Note
+    /// Only **transformation objects** (e.g. those created by calling [`new()`](fn@Proj::new())) can be
```
michaelkirk, line 510: If you do the wrong thing and try to create coordinate metadata for a pipeline, than [you'll hit this](https://github.com/georust/proj/discussions/220#discussioncomment-12833479), right?: 

> called `Result::unwrap()` on an `Err` value: ProjError("Unknown error (code 4096)")


Should we try to be a "better proj than proj" and encode this in the type system?

I'm ok either way for now.

michaelkirk (approved):
This seems important and I'd be glad to see it merged! 

It's a little rough around the edges, but if I understand correctly, that seems due to the underlying proj implementation.

src/proj.rs
urschrei: "Just" floats of the form year.month,and month can be 0: 2010.3 and 2014.0 are both valid values.

michaelkirk: I couldn't find it documented in proj, but in gdal, which I'm comfortable assuming has the same semantics:

https://gdal.org/en/stable/api/ogrspatialref.html#_CPPv4NK19OGRSpatialReference18GetCoordinateEpochEv

> coordinateEpoch Coordinate epoch as decimal year (e.g. 2021.3), or 0 if not set, or relevant.

So 2021.3 is April 2021.

src/proj.rs
urschrei: I played around with that idea a bit last night. The fundamental issue is that _we_ don't distinguish between a pipeline and a standard transformation object (neither does libproj), which means we can't get fancy with a bound (I don't think…): everything's a Proj. We have some runtime options:
1. Catch the error and return something more informative;
2. Check whether the input is a pipeline using proj_info, and return an informative error if kind is Some("Pipeline").

michaelkirk: Having a runtime check like you described could be a good compromise. We could at least produce a useful error message.

urschrei: I am very much open to something compile-time. But it would be very much a breaking change, right?

michaelkirk: Oh hell yes. We could maybe minimize it by having a default? But definitely breaking.

michaelkirk: I'm also fine to not do it (or do it later).

urschrei: OK, we've got a more informative error message. I think rather than breaking everyone for the sake of an operation that isn't required by most users, this will do for now.

PR #228, opened 2025-04-16
Update network (grid download) functionality to use ureq 3.x

urschrei:
---

<s>No changelog entry yet since we have other PRs in flight.</s>

This was an unfun piece of work as the ureq API has evolved significantly between 2.x and 3.0, and the way we're passing the connection pool object around is more difficult to do now. That having been said the tests pass and none of the changes affect the public API. We also have a new optional dep in the form of `http`, but I suspect it was present previously anyway.

### DIFF
--- a/src/network.rs
+++ b/src/network.rs
@@ -20,7 +20,7 @@ use std::io::Read;
 use std::ops::Range;
 use std::os::raw::c_ulonglong;
 use std::ptr::{self, NonNull};
-use ureq::{Agent, Request, Response};
+use ureq::Agent;
 
 use crate::proj::{ProjError, _string};
 use libc::c_char;
@@ -63,10 +63,10 @@ impl Drop for HandleData {
     }
 }
 
-/// Return an exponential wait time based on the number of retries
+/// Return a quadratically-increasing wait time based on the number of retries
 ///
 /// Example: a value of 8 allows up to 6400 ms of retry delay, for a cumulative total of 25500 ms
-fn get_wait_time_exp(retrycount: i32) -> u64 {
+fn get_wait_time_quad(retrycount: i32) -> u64 {
     if retrycount == 0 {
         return 0;
     }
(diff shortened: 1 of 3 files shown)


urschrei:
I don't know why tests are only failing with that dependency mismatch on 1.82, but incorporating #229 causes this PR to compile for me locally with `--features network` using rustc 1.82

src/network.rs
```diff
@@ -75,34 +75,46 @@ fn get_wait_time_exp(retrycount: i32) -> u64 {
...
+        while (SERVER_ERROR_CODES.contains(&res.status().as_u16())
+            || RETRY_CODES.contains(&res.status().as_u16()))
             && retries <= MAX_RETRIES
         {
             retries += 1;
             let wait = time::Duration::from_millis(get_wait_time_exp(retries as i32));
             thread::sleep(wait);
-            let retry = rb.clone();
-            *res = retry.call()?;
+            let agent = Agent::new_with_defaults();
```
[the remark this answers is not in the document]
urschrei: Somewhat sensible. Done in 85f8fc9.

src/network.rs
```diff
@@ -172,44 +184,62 @@ unsafe fn _network_open(
...
+        .headers()
+        .get("Content-Length")
+        .and_then(|val| val.to_str().ok())
+        .and_then(|s| s.parse::<usize>().ok())
+        .ok_or(ProjError::ContentLength)?;
+
     let headers = res
-        .headers_names()
-        .into_iter()
+        .headers()
```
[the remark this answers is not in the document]
urschrei: Not really worth it imo

src/network.rs
```diff
@@ -172,44 +184,62 @@ unsafe fn _network_open(
...
+        .keys()
```
lnicola, line 211: Here we're iterating over the keys and looking up the values again in the `filter_map` below.

src/network.rs
```diff
@@ -334,41 +364,58 @@ fn _network_read_range(
...
+    let headers = [("Range", hvalue.as_str()), ("Client", CLIENT)];
+
+    // hand the response off to the error-handler, continue on success
+    error_handler(&mut res, &hd.url, &headers, clt.clone())?;
+
     let headers = res
-        .headers_names()
-        .into_iter()
+        .headers()
+        .keys()
```
lnicola, line 383: Ditto

src/network.rs
```diff
@@ -75,34 +75,46 @@ fn get_wait_time_exp(retrycount: i32) -> u64 {
...
+        || RETRY_CODES.contains(&res.status().as_u16())
+    {
         // Start retrying: up to MAX_RETRIES
-        while (SERVER_ERROR_CODES.contains(&res.status()) || RETRY_CODES.contains(&res.status()))
+        while (SERVER_ERROR_CODES.contains(&res.status().as_u16())
+            || RETRY_CODES.contains(&res.status().as_u16()))
             && retries <= MAX_RETRIES
         {
             retries += 1;
             let wait = time::Duration::from_millis(get_wait_time_exp(retries as i32));
```
lnicola, line 95: Unrelated, but exponential back-off (a very popular approach) picks a random time between 0 and an exponentially-increasing limit.

lnicola (approved):
Looks reasonable overall.

src/network.rs
urschrei: Good catches. Simplified with an iterator now.

src/network.rs
urschrei: Ha fair. Rather than "fix" it, I've renamed it to `_quad` and updated the docstring.

lnicola: Oh wow, I didn't even notice it's `retries^2`, not `2^retries`.

PR #233, opened 2025-05-09
Adds an as_wkt() method on Proj

PaulWagener:
- [X] I agree to follow the project's [code of conduct](https://github.com/georust/.github/blob/main/CODE_OF_CONDUCT.md).
---

This adds support for [as_wkt](https://proj.org/en/stable/development/reference/functions.html#c.proj_as_wkt) method and all its options

### DIFF
--- a/src/proj.rs
+++ b/src/proj.rs
@@ -12,7 +12,7 @@ use proj_sys::{
     proj_trans, proj_trans_array, proj_trans_bounds, PJconsts, PJ_AREA, PJ_CONTEXT, PJ_COORD,
     PJ_DIRECTION_PJ_FWD, PJ_DIRECTION_PJ_INV, PJ_INFO, PJ_LPZT, PJ_WKT_TYPE_PJ_WKT1_ESRI,
     PJ_WKT_TYPE_PJ_WKT1_GDAL, PJ_WKT_TYPE_PJ_WKT2_2015, PJ_WKT_TYPE_PJ_WKT2_2015_SIMPLIFIED,
-    PJ_WKT_TYPE_PJ_WKT2_2019, PJ_XYZT,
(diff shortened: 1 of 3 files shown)


urschrei (approved):

src/proj.rs
```diff
@@ -1286,6 +1288,142 @@ impl Proj {
...
+                    if allow_linunit_node { "YES" } else { "NO" }
+                ))?);
+            }
+
+            if opts.is_empty() {
+                None
+            } else {
+                Some(opts)
+            }
+        } else {
```
michaelkirk, line 1356: Only a style nit, so take it or leave it.

Rather than having this far away dangling else, you could:
```
let options_str = options.as_ref().map(|options| {
   ...
});
```

src/proj.rs
```diff
@@ -1286,6 +1288,142 @@ impl Proj {
...
+}
+
+pub enum WktOutputAxis {
+    Auto,
+    Yes,
+    No,
+}
+
+#[allow(non_camel_case_types)]
+pub enum WktVersion {
```
michaelkirk, line 1421: It seems like there are some missing entries here: https://proj.org/en/stable/development/reference/datatypes.html#c.PJ_WKT_TYPE

I'm not familiar enough to know if these will be missed.

src/proj.rs
PaulWagener: There are question mark operators on the CString::new() method that won't return if I wrap the code in a closure. Atleast not without adding a transpose() on the end. I'd be happy to implement that if you prefer.

src/proj.rs
michaelkirk: Sounds reasonable to me. 👍

michaelkirk (approved):

PR #237, opened 2025-07-11
Bump libproj to 9.6.2, MSRV 1.85, 2024 edition

urschrei:
---

This updates:

- proj-sys to use libproj 9.6.2, MSRV 1.85, and the 2024 edition.
- proj to use proj-sys v0.27, and the MSRV and editions above

This PR is gated on images from https://github.com/georust/docker-images/pull/45 being available.


michaelkirk (approved):
Sorry this sat here for a million years!

PR #245, opened 2026-01-30
Add bundled_proj_sqlite feature to support windows-msvc

yutannihilation:
- [ ] I added an entry to the project's change log file if knowledge of this change could be valuable to users.
---

In order to make proj-sys work on Windows (more specifically, `windows-msvc`. I don't know well about `windows-gnu`), I found there are at least two options.

1. Use vcpkg to install PROJ
2. Bundle PROJ including SQLite

(I don't describe the details about first option, but here's what it looks like: https://github.com/MIERUNE/point-tiler/pull/64)

For the second option, the proj crate and proj-sys crate already exposes `bundled_proj`, but I found it's not enough on Windows' case, where we usually doesn't have SQLite installed. So, this pull request adds a feature flag `bundled_proj_sqlite` to expose the libsqlite3-sys's `bundled` feature in order to bundle SQLite as well.

That's all of this pull request, but I have to explain about some tricky things:

- PROJ uses the SQLite executable as well as the library (cf. <https://proj.org/en/stable/install.html#build-requirements>). So, even if we use `bundled_proj_sqlite`, we need to install SQLite on the GHA CI.
- For some reason, `cargo test` requires some Windows system libraries, otherwise it fails with the following error. So, I added these in `build.rs`.  
  ```
  = note: libproj_sys-119818f7f11b3538.rlib(filemanager.obj) : error LNK2019: unresolved external symbol __imp_CoTaskMemFree referenced in function proj_context_get_user_writable_directory␍
          libproj_sys-119818f7f11b3538.rlib(filemanager.obj) : error LNK2019: unresolved external symbol SHGetKnownFolderPath referenced in function proj_context_get_user_writable_directory␍
  ```
- ~~The `proj::test::test_inverse_projection` test reliably fails with `STATUS_ACCESS_VIOLATION`. I'm yet to figure out what this error is.~~ (fixed now)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -45,6 +45,7 @@ jobs:
       - proj-sys-ubuntu
       - proj-macos
       - proj-sys-macos
+      - proj-windows
     steps:
       - name: Mark the job as a success
         run: exit 0
@@ -58,6 +59,7 @@ jobs:
       - proj-sys-ubuntu
       - proj-macos
       - proj-sys-macos
+      - proj-windows
     steps:
       - name: Mark the job as a failure
         run: exit 1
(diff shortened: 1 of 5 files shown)


yutannihilation:
Thanks, I removed the commit from this pull request and pushed to https://github.com/georust/proj/pull/246

michaelkirk:
I don't personally have a way to test this, but it seems reasonable. Any concerns @urschrei?

urschrei:
Same re: testing, but I'm happy! Sorry we let this slide @yutannihilation.

yutannihilation:
No hurry at all! I understand it's tricky to review the code that works on platforms that you don't use. Please let me know if there's anything I can help.

PR #251, opened 2026-02-11
persist network client

### LINKED ISSUE #206: The network handler should reuse the HTTP client
`_network_open` should store the HTTP client, so further requests don't need to establish new connections.

Spun off #205, CC @TomFryersMidsummer just in case.

michaelkirk:
---

Based on #243 (though I rebased it), so merge that first.

Fixes #206

### DIFF
--- a/src/network.rs
+++ b/src/network.rs
@@ -165,9 +165,11 @@ pub(crate) unsafe extern "C" fn network_open(
         Err(e) => {
             let err_string = e.to_string();
             unsafe {
-                out_error_string
-                    .copy_from_nonoverlapping(err_string.as_ptr().cast(), err_string.len());
-                out_error_string.add(err_string.len()).write(0);
+                let len = err_string
+                    .len()
+                    .min(error_string_max_size.saturating_sub(1));
+                out_error_string.copy_from_nonoverlapping(err_string.as_ptr().cast(), len);
+                out_error_string.add(len).write(0);
             }
             ptr::null_mut()
         }
@@ -354,11 +356,13 @@ pub(crate) unsafe extern "C" fn network_read_range(
         Err(e) => {
             // The assumption here is that if 0 is returned, whatever error is in out_error_string is displayed by libproj
             // since this isn't a conversion using CString, nul chars must be manually stripped
-            let err_string = e.to_string().replace('0', "nought");
+            let err_string = e.to_string().replace('\0', "");
             unsafe {
-                out_error_string
-                    .copy_from_nonoverlapping(err_string.as_ptr().cast(), err_string.len());
-                out_error_string.add(err_string.len()).write(0);
+                let len = err_string
+                    .len()
+                    .min(error_string_max_size.saturating_sub(1));
+                out_error_string.copy_from_nonoverlapping(err_string.as_ptr().cast(), len);
+                out_error_string.add(len).write(0);
             }
             0usize
         }
--- a/src/network.rs
+++ b/src/network.rs
@@ -1,13 +1,13 @@
 #![deny(
     clippy::cast_slice_from_raw_parts,
     clippy::cast_slice_different_sizes,
-    clippy::invalid_null_arguments,
+    invalid_null_arguments,
     clippy::ptr_as_ptr,
     clippy::transmute_ptr_to_ref
 )]
-/// A module for native grid network functionality, so we don't have to depend on libcurl
-/// The crate-public functions are facades – they're designed for interaction with libproj –
-/// delegating actual functionality to non-public versions, prefixed by an underscore.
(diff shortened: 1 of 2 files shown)


src/network.rs
```diff
@@ -133,190 +179,79 @@ fn error_handler<'a>(
...
-    let cstr = CString::new(&**hvalue).unwrap();
-    let header = cstr.into_raw();
-    // Raw pointers are Copy: the pointer returned by this function is never returned by libproj so
-    // in order to avoid a memory leak the pointer is copied and stored in the HandleData struct,
-    // which is dropped when close_network returns. As part of that drop, all pointers in hptrs are freed.
-    hd.hptrs.push(
-        NonNull::new(header).expect("Failed to create non-Null pointer when building header value"),
-    );
-    Ok(header)
+    network_client
```
michaelkirk, line 250: Rather than storing `String`s and converting here, I updated the network request logic to persist the `CString`s, which makes this code a little simper.

src/network.rs
```diff
@@ -328,115 +263,41 @@ unsafe fn _network_get_header_value(
...
+        Ok(read) => {
+            unsafe {
+                *out_error_string = 0;
+            }
+            read
+        }
         Err(e) => {
-            // The assumption here is that if 0 is returned, whatever error is in out_error_string is displayed by libproj
-            // since this isn't a conversion using CString, nul chars must be manually stripped
-            let err_string = e.to_string().replace('0', "nought");
```
michaelkirk: Don't miss this! I got rid of the \0 handling. We should probably figure it out, but maybe better to do so in the context of #243 first, and then I can rebase.

src/network.rs
```diff
@@ -36,55 +36,106 @@ const SUCCESS_ERROR_CODES: Range<u16> = 200..300;
...
-impl Drop for HandleData {
-    // Whenever HandleData is dropped we free all allocated header pointers
-    fn drop(&mut self) {
-        for header in self.hptrs.drain(..) {
-            let _ = unsafe { CString::from_raw(header.as_ptr().cast()) };
-        }
+    /// Perform an HTTP range request starting at `offset` for `output_buffer.len()` bytes.
+    /// The response body is written into `output_buffer` and the response headers are
+    /// cached in `most_recent_response_headers`. Returns the number of bytes read.
+    fn read(&mut self, offset: u64, output_buffer: &mut [u8]) -> Result<usize, ProjError> {
```
michaelkirk, line 77: This `read` method DRYs up `network_open` and `network_read` a bit.

src/network.rs
```diff
@@ -36,55 +36,106 @@ const SUCCESS_ERROR_CODES: Range<u16> = 200..300;
...
-    res: &'a mut http::Response<ureq::Body>,
+fn request_with_retries(
     url: &str,
-    headers: &[(&str, &str)],
-    clt: Agent,
-) -> Result<&'a mut http::Response<ureq::Body>, ProjError> {
+    mut make_request: impl FnMut() -> Result<http::Response<ureq::Body>, ProjError>,
+) -> Result<http::Response<ureq::Body>, ProjError> {
     let mut retries = 0;
+    let mut res = make_request()?;
```
michaelkirk, line 138: Slightly tweaked this method - it now makes the initial request as well as potential retry requests. I think it flows a little better.

src/network.rs
```diff
@@ -133,190 +179,79 @@ fn error_handler<'a>(
...
-    };
-    match result {
-        Ok(res) => res,
-        #[allow(clippy::ptr_as_ptr)]
+    // Given a start and a length, we can create a rust slice
+    let output_buffer =
+        unsafe { std::slice::from_raw_parts_mut(buffer.cast::<u8>(), size_to_read) };
+
+    match unsafe { _string(url) }
+        .map_err(ProjError::from)
```
michaelkirk, line 198: Error handling in this method is a little clunky. The facade pattern was nicer for this, but I estimated it wasn't worth it for these 3 lines of error handling.

src/network.rs
```diff
@@ -133,190 +179,79 @@ fn error_handler<'a>(
...
+        .and_then(|url| {
+            NetworkClient::open(url, offset, output_buffer, unsafe { &mut *out_size_read })
+        }) {
+        Ok(network_client) => {
+            // clear out any error message when successful
+            unsafe {
+                *out_error_string = 0;
```
michaelkirk, line 205: I think this is equivalent to the previous:
```
    let err_string = "";
    unsafe {
        out_error_string.copy_from_nonoverlapping(err_string.as_ptr().cast(), err_string.len());
        out_error_string.add(err_string.len()).write(0);
    }
```

src/network.rs
urschrei: Resolved by d5a72f3c

urschrei (approved):

Issue #254, opened 2026-06-04
Compare Proj instances for equivalence


tmillenaar:
Hi

I would like to be able to check whether two Proj instances are compatible or whether some kind of resampling/conversion is needed. For this I would like to be able to do something like `proj1 == proj2` or `proj1.is_equivalent_to(proj2)`. It seems that currently there is no such method available. I could compare the definition strings, but I believe this is not necessarily stable as the .dev() field returns an optional. I could compare the wkt representation, but since the same CRS can be defined in multiple ways, this is not necessarily reliable either.
It seems that there are already exposed bindings for [proj_is_equivalent_to_with_ctx](https://github.com/georust/proj/blob/af8d787e563630a7922c497592eaef72fac246ee/proj-sys/src/bindings_docs-rs.rs#L1531).  As a user I do not think I am able to use this since I do not have access to the raw C pointers and context (correct me if I in fact can access to these). Could we add an equivalence function that makes use of this binding? I am willing to implement this myself if you think this would be beneficial to have.

Cheers,
Timo

urschrei:
Hi Timo,

As our `-sys` bindings expose it, you should be able to write a user-facing implementation in this crate; you'll have access to the context object in the crate, although I'm not sure what `PJ_COMPARISON_CRITERION` is. Have a look at how we've written some of the existing user-facing methods and let us know if you need help.

tmillenaar:
Hi urschrei, thanks for the fast reply!

I created the following PR: https://github.com/georust/proj/pull/255

There I also touch on what [PJ_COMPARISON_CRITERION](https://github.com/OSGeo/PROJ/blob/4c7ea9dd514069314417c3db2dde89ba5b8127a8/src/proj.h#L913) is about.

I am totally willing to adjust the implementation, suggestions are welcome

tmillenaar:
Implemented in PR: https://github.com/georust/proj/pull/255

PR #255, opened 2026-06-04
Add equivalence check

tmillenaar:
---

Addresses https://github.com/georust/proj/issues/254

`PJ_COMPARISON_CRITERION` is a C-enum [[see here](https://github.com/OSGeo/PROJ/blob/4c7ea9dd514069314417c3db2dde89ba5b8127a8/src/proj.h#L913)] describing the options to either be very strict, allow for semantic differences or allow for semantic difference and axis order differences. That makes three options, but since there is also a version with and without context, we have 6 versions in the original PROJ to compare.
In this MR I just went with a strict and loose version, both with context. While I feel this should be sufficient for most cases, if the other versions should also be supported, please comment so below. 

Also, now these are just methods on the Struct. It could be neat to implement one of these as PartialEq. That does do away with the equal/equivalent nuance of course, but would mean it could be used with `Vec::contains` for example. Would that be desirable?

### DIFF
--- a/src/proj.rs
+++ b/src/proj.rs
@@ -2,17 +2,20 @@ use libc::c_int;
 use libc::{c_char, c_double};
 use num_traits::Float;
 use proj_sys::{
-    PJ_AREA, PJ_CONTEXT, PJ_COORD, PJ_DIRECTION_PJ_FWD, PJ_DIRECTION_PJ_INV, PJ_INFO, PJ_LPZT,
-    PJ_WKT_TYPE_PJ_WKT1_ESRI, PJ_WKT_TYPE_PJ_WKT1_GDAL, PJ_WKT_TYPE_PJ_WKT2_2015,
-    PJ_WKT_TYPE_PJ_WKT2_2015_SIMPLIFIED, PJ_WKT_TYPE_PJ_WKT2_2019,
+    PJ_AREA, PJ_COMPARISON_CRITERION_PJ_COMP_EQUIVALENT,
+    PJ_COMPARISON_CRITERION_PJ_COMP_EQUIVALENT_EXCEPT_AXIS_ORDER_GEOGCRS,
+    PJ_COMPARISON_CRITERION_PJ_COMP_STRICT, PJ_CONTEXT, PJ_COORD, PJ_DIRECTION_PJ_FWD,
+    PJ_DIRECTION_PJ_INV, PJ_INFO, PJ_LPZT, PJ_WKT_TYPE_PJ_WKT1_ESRI, PJ_WKT_TYPE_PJ_WKT1_GDAL,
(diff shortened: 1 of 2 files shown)


urschrei:
I think this is pretty good. I like your `PartialEq` idea, but how about only providing it for `is_equal_to`, since it doesn't seem like it's a good fit for `equivalent`? Up to you, since you're presumably going to be using the feature.

src/proj.rs
```diff
@@ -1343,6 +1346,37 @@ impl Proj {
             Ok(_string(wkt)?)
         }
     }
+
+    /// Checks whether two proj instances are identical in all respects.
+    ///
+    /// For a less strict comparison, see is_equivalent_to
+    pub fn is_equal_to(&self, other: &Proj) -> bool {
```
michaelkirk: I think from an english language perspective, equal vs equivalent are too similar to be useful method names. 

Rather than relying on the user to read the docs every time they call the method, how about exposing what we actually mean?

Something like:
```
proj_1.is_equivalent_to(proj_2, ComparisonCriterion::Strict)
proj_1.is_equivalent_to(proj_2, ComparisonCriterion::AxisOrder)
```

IMO that better matches the proj API without inventing our own vocabulary.

michaelkirk (commented):
I'm willing to be overruled by @urschrei if he disagrees, but I'm a skeptical of adding a `PartialEq` implementation because there are multiple definitions of "equalish" with Proj in particular. Given it's potential ambiguity, I'd err on the side of forcing people to be explicit with equality checks.

Doing `contains` the explicit way might look like this:
```
let proj_1 = todo!();

projs.iter().any(|x| proj_1.is_equivalent_to(x, ComparisonCriteria::Strict))
projs.iter().any(|x| proj_1.is_equivalent_to(x, ComparisonCriteria::AxisOrdering))
```

It's longer, but clearer in my opinion.

src/proj.rs
michaelkirk: As another idiomatic style nit, rust typically omits the `is_` portion of boolean queries. 

In the same way that `2.eq(1 + 1)`  (**not** `2.is_eq(1 + 1)`), I'd prefer `proj_1.equivalent_to(proj_2, ComparisonCriterion::Strict)`.

tmillenaar: Fair enough, I collapsed it back into a single 'equivalent_to' method with an enum for the comparison_criterion. If we are using an enum now for this anyway, I might as well support all three options of the original implementation.

tmillenaar:
@michaelkirk 
I think the current state of this MR aligns with your preferences. Do you think we can move ahead? Or have I left something unaddressed?

michaelkirk (approved):

michaelkirk:
Thanks for the follow up! I'll wait another day before merging in case @urschrei wants to weigh in.

urschrei (approved):
It'll need a rebase, but otherwise all good.

tmillenaar:
Great, thanks guys. Rebased.

urschrei:
Sorry, should have said: please also squash these into a single commit.

tmillenaar:
Alright, now squashed as well ;)

Issue #256, opened 2026-06-09
Performance of recreating the proj context


tmillenaar:
Hey all,

I am using this crate for an app that renders in real time. Big transformations are of course done asynchronously, but to my mild surprise I also had to be careful when creating a new Proj instance. I was creating several new Proj instances every frame, but that took between 8-9ms per instance. Since I only have 16ms in a frame if I aim at 60fps, creating new Proj instances every frame is a no-go. I now work around it by implementing my own crs cache where I wrap every Proj instance in an Arc. This does the job just fine, but this surely is not the most ergonomic way to use the crate. If you want to verify the timing, you can add the following test in proj.rs:

```
    #[test]
    fn bench_proj_creation_time() {
        let n = 1000;

        // Warmup (would avoid first-run initialization)
        for _ in 0..100 {
            let _ = Proj::new("EPSG:4326").unwrap();
        }

        let start = Instant::now();

        for _ in 0..n {
            let _ = Proj::new("EPSG:4326").unwrap();
        }

        let elapsed = start.elapsed();

        let avg = elapsed / n;

        println!("Created {} Proj instances", n);
        println!("Total time: {:?}", elapsed);
        println!("Mean per Proj::new: {:?}", avg);
    }
```

Then run it like so:
cargo test proj::test::bench_proj_creation_time -- --nocapture

I then get:
```
Created 1000 Proj instances
Total time: 8.821968393s
Mean per Proj::new: 8.821968ms
```

In contrast, in C I get a median time of 1.4 us and in python (!) I even get a mere ~7us. I can provide those scripts as well if you so desire.

I think the big difference is that in this crate we re-create the context every time we instantiate a new Proj object:
https://github.com/georust/proj/blob/f0ab553f59ef83862ceaefcdfcdd0add7446f219/src/proj.rs#L704

I imagine that every Proj instance owning it's own context is quite convenient and secure, but seeing both the Python and C implementations avoid this makes me wonder if we cannot do the same. Is there a particular reason we do it this way? Or is it mostly a convenience thing? I imagine we could maybe utilize a OnceCell or something like it, but before I dive too deep into the rabbit hole, maybe you guys can cut my investigation short if there is a known dealbreaker or deliberate design decision that would prevent some kind of shared context.

I would love to hear your thougts.

Kind regards,
Timo Millenaar

urschrei:
Hi @tmillenaar, have a look at https://github.com/georust/proj/pull/257.

tmillenaar:
[...]

PR #257, opened 2026-06-10
Don't call proj_cleanup() on every Drop

urschrei:
proj_cleanup() frees PROJ's process-global caches (grids, +init files), so calling it on each Proj/ProjBuilder drop clears caches the next object would reuse, forcing reloads from disk. Removing it drops Proj::new cost by ~10x

---

See #256. We're still a long way off c / cpp perf, but this is a minor intervention for a lot of perf.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -137,6 +137,19 @@
 //! The path used to search for resource files can be modified using
 //! [`set_search_paths`](proj/struct.ProjBuilder.html#method.set_search_paths)
 //!
+//! ### Resource Caching
+//! Separately from the on-disk [Grid File Cache](#grid-file-cache) described above, PROJ keeps
+//! process-global caches of resources *in memory*: grids that have been loaded, and the contents
+//! of `+init` files. These caches are shared across all `Proj` and `ProjBuilder` instances and
+//! persist for the lifetime of the process; they are not released when individual instances are
+//! dropped. Repeated creation of transformation objects therefore reuses already-loaded resources
+//! rather than reading them from disk again.
+//!
+//! These in-memory caches are never explicitly freed, so they remain resident until the process
+//! exits, at which point the operating system reclaims them. They are bounded by the set of
+//! resources actually loaded and so do not grow without bound: the cost is memory residency for
+//! the lifetime of the process, not additional disk usage.
+//!
 //! ## Conform your own types
 //!
 //! If you have your own geometric types, you can conform them to the `Coord` trait and use `proj`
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -137,6 +137,13 @@
 //! The path used to search for resource files can be modified using
 //! [`set_search_paths`](proj/struct.ProjBuilder.html#method.set_search_paths)
 //!
+//! ### Resource Caching
+//! PROJ maintains process-global caches of resources such as grid files and the contents of
+//! `+init` files. These caches are shared across all `Proj` and `ProjBuilder` instances and
+//! persist for the lifetime of the process: they are not released when individual instances are
```
michaelkirk: > persist for the lifetime of the process

Now that we've removed the per-instance cleanup, what mechanism cleans them up at the end of the process?

urschrei: Ah! I should have said: nothing. The OS reclaims it when the process terminates. The docs say:

>  "This function frees global resources (grids, cache of +init files). It should be called typically before process termination, and after having freed PJ and PJ_CONTEXT objects."

So it doesn't have to be called, in my reading. Pyproj doesn't ([or at least, I couldn't find any instance of `proj_cleanup`](https://github.com/search?q=repo%3Apyproj4%2Fpyproj%20proj_cleanup&type=code)), FWIW.

michaelkirk: We should at minimum document that more clearly. I guess the risk is that you're taking up disk space. 

Additionally, since we're no longer cleaning up automatically, what would you think about adding something like a `Proj::proj_cleanup()` method in the (admittedly unlikely) case that users are relying on this, so that they have a way they can manually achieve the previously automatic behavior?

urschrei: > I guess the risk is that you're taking up disk space

The cache in question is the in-memory cache, not disk. But yep, I can add a line about that.

On the cleanup method, we can add it, but it should probably be a free function, since it's only meant to be called **after** every `PJ` / `PJ_CONTEXT` is destroyed, so having it on `self` would be bad.

michaelkirk: > The cache in question is the in-memory cache, not disk. But yep, I can add a line about that.

I misunderstood - sorry! 

I thought this was some kind of persistent cache. It's not obvious to me why you should call this before process termination, if it's only clearing a bunch of in-memory objects. They'd be reclaimed once the process terminates anyway, right?

michaelkirk (approved):
This is all good, and any concerns were just my own misunderstanding.

src/lib.rs
urschrei: > They'd be reclaimed once the process terminates anyway, right?

Yep!

PR #258, opened 2026-06-10
Add context type

urschrei:
---

<s>Based on #257, so merge that first.</s> This is part 1 of a 2-part stack which improves perf of `Proj::new` creation to around 1us per instance.

### DIFF
--- a/src/proj.rs
+++ b/src/proj.rs
@@ -8,12 +8,12 @@ use proj_sys::{
     PJ_DIRECTION_PJ_INV, PJ_INFO, PJ_LPZT, PJ_WKT_TYPE_PJ_WKT1_ESRI, PJ_WKT_TYPE_PJ_WKT1_GDAL,
     PJ_WKT_TYPE_PJ_WKT2_2015, PJ_WKT_TYPE_PJ_WKT2_2015_SIMPLIFIED, PJ_WKT_TYPE_PJ_WKT2_2019,
     PJ_WKT_TYPE_PJ_WKT2_2019_SIMPLIFIED, PJ_XYZT, PJconsts, proj_area_create, proj_area_destroy,
-    proj_area_set_bbox, proj_as_projjson, proj_as_wkt, proj_context_clone, proj_context_create,
-    proj_context_destroy, proj_context_errno, proj_context_get_url_endpoint,
-    proj_context_is_network_enabled, proj_context_set_search_paths, proj_context_set_url_endpoint,
-    proj_coordinate_metadata_create, proj_coordinate_metadata_get_epoch, proj_create,
-    proj_create_crs_to_crs, proj_create_crs_to_crs_from_pj, proj_destroy, proj_errno_string,
-    proj_get_area_of_use, proj_grid_cache_set_enable, proj_info, proj_is_equivalent_to_with_ctx,
+    proj_area_set_bbox, proj_as_projjson, proj_as_wkt, proj_context_errno,
+    proj_context_get_url_endpoint, proj_context_is_network_enabled, proj_context_set_search_paths,
+    proj_context_set_url_endpoint, proj_coordinate_metadata_create,
+    proj_coordinate_metadata_get_epoch, proj_create, proj_create_crs_to_crs,
+    proj_create_crs_to_crs_from_pj, proj_destroy, proj_errno_string, proj_get_area_of_use,
+    proj_grid_cache_set_enable, proj_info, proj_is_equivalent_to_with_ctx,
     proj_normalize_for_visualization, proj_pj_info, proj_trans, proj_trans_array,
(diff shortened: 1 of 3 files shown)


src/proj.rs
```diff
@@ -217,11 +218,75 @@ fn area_set_bbox(parea: *mut proj_sys::PJ_AREA, new_area: Option<Area>) {
...
+            Context::Shared(shared) => shared.ctx,
+        }
+    }
+}
+
+impl Drop for Context {
+    fn drop(&mut self) {
+        // `Owned` contexts are destroyed here; `Shared` contexts are destroyed by
+        // `SharedContext::drop` once the last reference (thread-local or `Proj`) is gone.
+        if let Context::Owned(ctx) = *self {
```
michaelkirk: Rather than having two drop methods, it'd be better to have the Owned and Shared variants wrap a single type wrapping the `* PJ_CONTEXT`, and thus only have a single Drop impl.

src/proj.rs
```diff
@@ -217,11 +218,75 @@ fn area_set_bbox(parea: *mut proj_sys::PJ_AREA, new_area: Option<Area>) {
     }
 }
 
+/// A PROJ context (`PJ_CONTEXT`) owned or borrowed by a [`Proj`].
+///
+/// `Owned` contexts are created (or cloned) for a single `Proj` and destroyed when it is dropped.
+/// `Shared` contexts are the per-thread context returned by [`thread_local_context`]; the context
+/// is reference counted so that it outlives every `Proj` that uses it, regardless of the order in
+/// which the thread-local and any surviving `Proj` instances are dropped at thread exit.
+enum Context {
```
michaelkirk: This file is already really big - consider moving the new stable of `Context` functionality into their own module.

michaelkirk (approved):
The approach LGTM and seems like it should work great to speed things up.

I have some stylistic nits, which you should fee free to ignore:

In the proposed implementation, Context is a hidden concept. Do you think it's worth exposing to the user?

I think this could be a bit tidier if we fully embraced Context as a concept. For instance, I mentioned in line the unified Drop handler.

Also, we have some legacy free functions, which now "smell like" instance methods, which could make sense to hang on Context.

```
fn transform_string(ctx: Context...
fn transform_epsg(ctx: Context...
fn crs_to_crs_from_pj(ctx: Context...
```

If this feedback does not "spark joy" feel free to merge as is. I'd be happy to follow up myself now that I've thought about it.

michaelkirk:
Just to give you an idea of what I mean, something like:

```rust
// What you're currently calling `Context` - admittedly `ContextEnum` is a terrible name... 
// but I haven't thought hard about it yet. 
enum ContextEnum {
    Owned(Context),
    // Wired into Proj::new/new_known_crs in a follow-up commit.
    #[allow(dead_code)]
    Shared(Rc<Context>),
}

impl ContextEnum {
    fn as_ptr(&self) -> *mut PJ_CONTEXT {
        match self {
            ContextEnum::Owned(ctx) => ctx.as_ptr(),
            ContextEnum::Shared(shared) => shared.as_ptr(),
        }
    }
}

/// An Owned instance of the PJ_CONTEXT from libproj.
/// Upon being dropped, proj_context_destroy will be called on the PJ_CONTEXT
struct Context {
    ctx: *mut PJ_CONTEXT,
}

impl Context {
    fn new() -> Self {
        Self { ctx: unsafe { proj_context_create() } }
    }

    fn as_ptr(&self) -> *mut PJ_CONTEXT {
        self.ctx
    }
}

impl Clone for Context {
    fn clone(&self) -> Self {
        Self {
            ctx: unsafe { proj_context_clone(self.ctx) },
        }
    }
}

// No need for separate Shared vs. Owned Drop impls. 
impl Drop for Context {
    fn drop(&mut self) {
        // Must only destroy the context: this runs during thread-local teardown, so it must not
        // re-enter the `SHARED_CONTEXT` thread-local. We deliberately do not call proj_cleanup()
        // (see the note in `Drop for Proj`).
        unsafe { proj_context_destroy(self.ctx) };
    }
}
```

Admittedly, this might touch a lot of things, but I think it'd be a nice cleanup. We can do it separately from this perf work if that's your preference.

src/proj.rs
tmillenaar: This function is a recent addition to main. I take it you rebased recently, which means we should also use the new context syntax here.

```suggestion
                self.ctx(),
```

src/proj.rs
urschrei: Done (in new module)!

urschrei:
Those are sensible suggestions @michaelkirk. PTAL!

I haven't done anything about public-`Context` yet, but I think it does make sense. Let's merge this stack if we're happy and then build it.

michaelkirk (approved):
LGTM!

PR #259, opened 2026-06-10
Reuse context

urschrei:
---

Based on #258, so merge that first. This is part 2 of a 2-part stack which improves perf of `Proj::new` creation to around 1us per instance.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -150,6 +150,18 @@
 //! resources actually loaded and so do not grow without bound: the cost is memory residency for
 //! the lifetime of the process, not additional disk usage.
 //!
+//! ## Threading
+//!
+//! Each PROJ context (`PJ_CONTEXT`) is tied to a single thread and must not be used concurrently
+//! from more than one thread. `Proj` is therefore deliberately neither `Send` nor `Sync`; to
+//! transform coordinates on several threads, create a `Proj` on each thread.
+//!
+//! [`Proj::new`] and [`Proj::new_known_crs`] reuse one context per thread rather than creating a
+//! new one for every object. This keeps the connection to the PROJ database and its caches warm,
+//! so constructing many transformation objects on a thread is considerably cheaper than it would
+//! otherwise be. [`ProjBuilder`] keeps its own context, since it is used to configure
+//! context-specific state such as network access and search paths.
+//!
 //! ## Conform your own types
 //!
 //! If you have your own geometric types, you can conform them to the `Coord` trait and use `proj`
--- a/src/proj.rs
+++ b/src/proj.rs
@@ -23,7 +23,7 @@ use std::{
     str,
 };
 
-use crate::context::{Context, ProjContext};
+use crate::context::{Context, ProjContext, thread_local_context};
 use crate::cstring_array::CStringArray;
 
 #[cfg(feature = "network")]
@@ -673,7 +673,7 @@ impl Proj {
     // PJ_LP signals projection of geodetic coordinates, with output being PJ_XY
     // and vice versa, or using PJ_XY for conversion operations
     pub fn new(definition: &str) -> Result<Proj, ProjCreateError> {
-        ProjContext::Owned(Context::new()).transform_string(definition)
+        thread_local_context().transform_string(definition)
     }
 
     /// Try to create a new transformation object that is a pipeline between two known coordinate reference systems.
@@ -728,7 +728,7 @@ impl Proj {
         to: &str,
         area: Option<Area>,
     ) -> Result<Proj, ProjCreateError> {
-        ProjContext::Owned(Context::new()).transform_epsg(from, to, area)
+        thread_local_context().transform_epsg(from, to, area)
     }
 
     /// Create a transformation object that is a pipeline _between_ two known coordinate reference systems.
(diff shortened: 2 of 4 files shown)


src/proj.rs
```diff
@@ -756,8 +823,8 @@ impl Proj {
         options: Option<Vec<&str>>,
     ) -> Result<Proj, ProjCreateError> {
         // Clone the context to avoid double-free in Drop implementations
-        let ctx = unsafe { proj_context_clone(self.ctx) };
-        crs_to_crs_from_pj(ctx, self, target_crs, area, options)
+        let ctx = unsafe { proj_context_clone(self.ctx()) };
+        crs_to_crs_from_pj(Context::Owned(ctx), self, target_crs, area, options)
```
tmillenaar: Here `create_crs_to_crs_from_pj` gets an Owned Context whereas `new_known_crs` has a Shared Context. Is this difference deliberate?

urschrei: It's deliberate: the "derive from an existing `Proj`" methods (`create_crs_to_crs_from_pj` etc.) clone `self`'s context, whereas the from-scratch constructors (`new`, `new_known_crs`) use the per-thread shared context.

I think the `clone` originally existed only to avoid a double-free from sharing `self`'s raw context pointer which the ref-counted `Shared` context now handles anyway. The only remaining reason to clone is to inherit `self`'s context config (search paths / network endpoint / CA bundle) into the derived transformation, which only matters when `self` came from a configured `ProjBuilder`.

The **downside** is that the clone opens a fresh `proj.db` connection, so this path doesn't get the #256 speedup. If we don't care about config inheritance here I'm happy to switch it to `thread_local_context()` for consistency and perf; if we do, I'll document it.

urschrei:
Some more (hopefully useful) context, since I wanted to make sure that we were doing it the "correct" way now:

PROJ's recommended model is one context per thread, shared by all objects created on that thread, rather than one context per object. From the [official quickstart](https://proj.org/en/stable/development/quickstart.html):

> It is recommended to create one threading context per thread used by the program. This ensures that all PJ objects created in the same context will be sharing resources such as error-numbers and loaded grids.

This is the pattern used by GDAL (a thread-local `PJ_CONTEXT` in [`ogr_proj_p.cpp`](https://github.com/OSGeo/gdal/blob/master/ogr/ogr_proj_p.cpp)), MapServer ([MapServer#7161](https://github.com/MapServer/MapServer/pull/7161)), and pyproj since 3.7.0 ([pyproj#1419](https://github.com/pyproj4/pyproj/pull/1419)). I realise that pyproj isn't "official" but it's in wide use and the authors spend a lot of time discussing their approach with the proj maintainers.

src/proj.rs
tmillenaar: Yeah, the guard against the double free makes sense. Personally I tend to use `new_known_crs` over `create_crs_to_crs_from_pj` anyway because I appreciate the axis normalization that happens in `new_known_crs`. That this path does not get the speedup therefore would not affect me, though I can imagine it is nice have these methods more aligned in that sense. If however that breaks with the ProjBuilder path I understand some caution is warranted, especially if there are people out there count on the difference in context.

urschrei:
(Need someone to OK this so we can merge)

src/lib.rs
```diff
@@ -150,6 +150,18 @@
 //! resources actually loaded and so do not grow without bound: the cost is memory residency for
 //! the lifetime of the process, not additional disk usage.
 //!
+//! ## Threading
+//!
+//! Each PROJ context (`PJ_CONTEXT`) is tied to a single thread and must not be used concurrently
+//! from more than one thread. `Proj` is therefore deliberately neither `Send` nor `Sync`; to
```
michaelkirk, line 156: It seems like it could be `Send`? But anyway, it's an improvement from what we currently have.

This behavior of using a thread-local is going to be a big speedup for pretty much everybody, but I'm wondering if there is a niche use case where you want to manage context specifically. 

With the c-lib you can have multiple contexts per threads (or multiple threads per context if you're careful that they don't overlap access). You can't do that with the rust lib now, right? I think this is a minority (maybe verging on non-existent) use case, but want to make sure I'm understanding.

michaelkirk (approved):
LGTM - I had a question, but it's just for my own understanding.

src/lib.rs
urschrei: I don't think we could make the new design `Send` because of `ProjContext::Shared(Rc<Context>)`: `Rc`'s ref counts are non-atomic. So sending a `Shared` to another thread. Would keep a clone of the original `Rc` (or vice versa?), and that would allow the ref count to be mutated by two threads at once. That's certainly a data race, and might be UB (??).

I think we can do multiple `ProjContext`s per thread anyway: `ProjBuilder` keeps its own owned context, and `clone_owned()` hands a derived `Proj` an independent context. The thread-local sharing _only_ applies to `Proj::new` and `Proj::new_known_crs`. So if you want a dedicated context, you can go through `ProjBuilder`.

Multiple threads per one context is allowed by The C lib if you serialise access yourself, but our wrapper can't enforce (even before this PR) anything that would make that safe at compile time. Maybe worth thinking about, but I haven't…

urschrei: We could improve the docs with e.g. "If you need a context independent of the per-thread one, construct via `ProjBuilder`"

michaelkirk: That makes sense, thanks for explaining.

